PR #36, opened 2026-01-13
Use ambient-id

jku:
For comparison with #35: 
* This PR also adds support for ambient-id
* Tested with interop.yml on my fork and with a custom pipeline on GitLab 
* removes ambient.rs completely: sigstore-rust now takes no opinions on the supported CI systems etc
* oidc API changes slightly to allow both None and Err from ambient id detection

Feel free to close if you like 35 more


jku:
> untested

I'm planning to get to this... it's just other work keeps preventing playtime with CIs

PR #69, opened 2026-02-16
refactor: API for fetching / using the trust root

wolfv:
With this refactor we strongly prefer to use TUF for the sigstore trust root instead of the embedded one as we did before.


crates/sigstore-trust-root/src/tuf.rs
```diff
@@ -216,27 +374,55 @@ impl TufClient {
...
+                    match self.check_cache_expiration(&cache_dir).await {
+                        Ok(()) => return Ok(bytes),
+                        Err(e) => {
+                            tracing::warn!(
+                                "Cached TUF metadata has expired ({}), falling back to embedded data",
+                                e
+                            );
+                            // Fall through to embedded data
+                        }
+                    }
```
jku: I'm not sure about this code. Why would the embedded content suddenly be better if timestamp has expired?

I think there is an argument to be made that TUF cached content (when not verified by TUF) may not be as "trusted" as the embedded data  -- but that has nothing to do with timestamp expiry

jku (approved):
I've not tried this yet myself but the code looks fine to me, thanks for doing all the work.

* The main feature of supporting any TUF url looks good, will be useful immediately in root-signing
* I think it would be good if the default and most obvious offline mechanism was _"use TUF material but offline"_ (instead of, _"use embedded material"_) -- I'm not quite sure if this is the case now (and I know the TUF libraries might not make this as useful as it should be) so I don't think this PR needs to handle all of it

PR #80, opened 2026-04-15
Allow timestamp-backed bundles without tlog entries

jdx:
## Summary

Allow bundles without transparency-log entries to pass bundle validation when the caller explicitly disables inclusion-proof verification and the bundle carries RFC3161 timestamp verification data.

This supports v0.3 bundles returned by GitHub's artifact attestation API, which can include `timestampVerificationData.rfc3161Timestamps` while omitting Rekor transparency-log entries.

The default behavior remains strict: `validate_bundle` still requires tlog entries because `ValidationOptions::default()` keeps `require_inclusion_proof: true`.

## Spec check

This matches the Sigstore bundle protobuf spec:

- The current JSON bundle media type is `application/vnd.dev.sigstore.bundle.v0.3+json`.
- `TimestampVerificationData.rfc3161_timestamps` is explicitly described as usable when the entry has not been stored on a transparency log.
- `VerificationMaterial.tlog_entries` is a repeated field, not marked required.
- `VerificationMaterial.timestamp_verification_data` is a separate source of timestamp material. The spec also notes timestamp may come from a tlog signed entry timestamp, so timestamp and tlog are alternatives for establishing signing time.

Relevant source: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L38-L47 and https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto#L98-L108

This PR keeps the trust-model distinction intact:

- `require_inclusion_proof: true`: still rejects bundles with no tlog entries.
- `require_inclusion_proof: false` plus RFC3161 timestamp data: bundle validation can proceed.
- `require_inclusion_proof: false` with neither tlog entries nor RFC3161 timestamp data: still rejected.

`sigstore-verify` already verifies TSA timestamp data when determining certificate validation time before falling back to Rekor integrated time, so this removes an early structural-validation rejection and leaves timestamp verification to the existing verification path.

## GitHub API example

GitHub's docs for the attestation list endpoint state that consumers must cryptographically verify signatures and timestamps and validate signer identity: https://docs.github.com/en/rest/users/attestations#list-attestations

A current GitHub artifact attestation response for `jdx/communique@0.1.9` has a v0.3 bundle with no transparency-log entries and one RFC3161 timestamp.

```sh
curl -fsSL \
  https://github.com/jdx/communique/releases/download/v0.1.9/communique-x86_64-unknown-linux-gnu.tar.gz \
  -o communique.tar.gz

digest=$(sha256sum communique.tar.gz | awk '{print $1}')

gh api "/repos/jdx/communique/attestations/sha256:$digest" \
  -H 'Accept: application/vnd.github+json' \
  -H 'X-GitHub-Api-Version: 2022-11-28' \
  --jq '.attestations | length as $count | .[0].bundle as $b | {
    attestation_count: $count,
    mediaType: $b.mediaType,
    tlogEntries_len: ($b.verificationMaterial.tlogEntries // [] | length),
    rfc3161Timestamps_len: ($b.verificationMaterial.timestampVerificationData.rfc3161Timestamps // [] | length),
    verificationMaterial_keys: ($b.verificationMaterial | keys)
  }'
```

Observed output:

```json
{
  "attestation_count": 1,
  "mediaType": "application/vnd.dev.sigstore.bundle.v0.3+json",
  "tlogEntries_len": 0,
  "rfc3161Timestamps_len": 1,
  "verificationMaterial_keys": [
    "certificate",
    "timestampVerificationData"
  ]
}
```

Before this PR, `validate_common` rejected this shape before signature or timestamp verification, even when the caller had disabled tlog verification through policy. After this PR, that caller policy can proceed, while default validation remains tlog-strict.

## Tests

- `cargo test -p sigstore-bundle test_v03_bundle_with_timestamp_and_no_tlog_entries`
- `cargo test -p sigstore-bundle`


jdx:
@wolfv sorry if this is a lot of noise from my clanker, but I wanted to show the digging I did there. It does seem to be that this is a valid fix since the spec is quite clear that the tlog is optional:

https://docs.sigstore.dev/about/bundle/

> Transparency Log entries can provide proof that a signing event was written to a public log. They are not required by the spec but inclusion of one or more transparency log entries is highly encouraged for bundles subject to public consumption. The embedded signed entry timestamp may be used to validate signing occurred during ceritificate validity.

wolfv:
Thanks @jdx - taking a look now!

wolfv:
@jdx do you have a bundle straight from Github that we could add to test this?

wolfv (approved):

jdx:
that's what crates/sigstore-bundle/tests/fixtures/bundle_v3_jdx_communique.json is, it comes from https://api.github.com/repos/jdx/communique/attestations/sha256:b958c6046bab52febf958c94974e1ffcc450bff78c28d7233e179bfd73828912

jdx:
thanks so much @wolfv !

PR #83, opened 2026-04-23
fix(sigstore-tsa): honor SignerInfo digest algorithm in CMS verification

jdx:
## Summary

`verify_message_digest_attribute` hashed the `TSTInfo` content with SHA-256 unconditionally, regardless of what `SignerInfo.digestAlgorithm` (RFC 5652 §5.3) declared. That works for the public-good Sigstore TSA — which signs with SHA-256 — but rejects every TSA response that uses any other digest. The companion `verify_message_imprint` function one definition above already reads the digest algorithm from the message imprint and switches on SHA-256 / SHA-384 / SHA-512; this PR mirrors that pattern for the CMS message-digest attribute.

## Why

The concrete trigger is GitHub Actions's internal TSA at `timestamp.githubapp.com`, which signs CMS `SignedData` with SHA-384. Any GitHub-issued artifact attestation (the bundles served from `/repos/{owner}/{repo}/attestations/<digest>` for builds that publish via [`actions/attest-build-provenance`](https://github.com/actions/attest-build-provenance)) currently fails with:

```
HashMismatch { expected: <48 hex bytes>, actual: <32 hex bytes> }
```

48 hex bytes = SHA-384 (what the TSA signed); 32 hex bytes = SHA-256 (what `sigstore-tsa` re-computed). The `SignerInfo.digestAlgorithm` OID is already extracted four lines below in `verify_cms_signature` for ECDSA signature verification (`let digest_alg_oid = &signer_info.digest_alg.oid;`). This PR threads that same OID into `verify_message_digest_attribute` and uses it.

## Approach

| Before | After |
| --- | --- |
| `verify_message_digest_attribute(signed_attrs, tst_info_der)` | `verify_message_digest_attribute(signed_attrs, tst_info_der, digest_alg_oid)` |
| `digest(&SHA256, tst_info_der)` | `digest(&{SHA256\|SHA384\|SHA512}, tst_info_der)` based on the OID, with `Error::ParseError` for unknown OIDs |

The OID extraction in `verify_cms_signature` is moved up four lines so it can be passed into both `verify_message_digest_attribute` and the existing `verify_ecdsa_signature` call.

## Discovered in

[`mise`'s sigstore-rust migration](https://github.com/jdx/mise/pull/9260). Investigation thread: [comment](https://github.com/jdx/mise/pull/9260#issuecomment-4306552714). Three real GitHub-issued bundles that hit this:

- `jdx/communique@v0.1.9` *(used as the regression-test fixture below)*
- `jdx/fnox@1.21.0`
- `astral-sh/uv@0.11.7`

All three are v0.3 bundles, TSA-only (no `tlogEntries`), SHA-384.

## Test plan

New `test_verify_github_sha384_timestamp` uses the `jdx/communique@v0.1.9` bundle as a real-world fixture (committed under `test_data/timestamps/github_sha384_bundle.json`, 3.7 KB) plus GitHub's published TSA chain narrowed to `timestampAuthorities` (fetched once from <https://tuf-repo.github.com/> and committed under `test_data/timestamps/github_trusted_root.json`, 26 KB). The existing `test_verify_valid_timestamp` case still covers SHA-256 — this PR is purely additive in coverage.

- [x] `cargo test -p sigstore-tsa` — **15/15 pass** (1 new)
- [x] `cargo test` — full workspace passes
- [x] `cargo fmt --check`
- [x] `cargo clippy -p sigstore-tsa --all-targets`

## Compat / risk

Internal helper signature change (`verify_message_digest_attribute` is `fn`, not `pub`). No public API change. Behaviour change is strictly broadening — bundles that verified before still verify; bundles that previously failed *only* because of the SHA-256 hardcode now succeed; everything else continues to fail at the same step.


jku (approved):
looks good to me, thanks.

The tsa test data could just be timestamp responses+certificates but this fits the existing test so seems fine.

wolfv:
Thank you @jdx!

PR #85, opened 2026-04-29
feat: trust root update and dependency update

wolfv:
## Summary

  Update the embedded Sigstore TUF bootstrap roots and staging trust material.

  - Replace the production embedded TUF root with the current v14 root, matching sigstore-python and sigstore/root-signing
  - Replace the staging embedded TUF root with the current v14 root from sigstore/root-signing-staging
  - Refresh staging trusted root and signing config
  - Enable an explicit rustls TLS backend for tough’s reqwest 0.12 dependency so HTTPS TUF fetching does not depend on incidental feature unification

## Notes

  This removes the need for the custom production TUF root workaround: the embedded production bootstrap root is no longer the old v1 metadata with the offset `expires` timestamp.

## Verification

  - `cargo test -p sigstore-trust-root --locked`
  - `cargo fmt --check`
  - `git diff --check`
  - Online smoke test for `fetch_trust_material()` against production and staging TUF repositories

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,11 +24,11 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c1daa54020e05aa0b163ee10434fff35a0f18d28a1cafa142bd1290e1abe630e"
 dependencies = [
  "astral-reqwest-middleware",
- "reqwest 0.13.2",
+ "reqwest 0.13.3",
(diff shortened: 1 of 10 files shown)


jku (approved):
Makes sense
* the feature stuff I am not familiar with but looks logical
* the embedded tuf update is good: I'll file a bug to automate that

PR #88, opened 2026-04-30
feat: add GitHub artifact attestation trusted root

wolfv:
#### Summary

Adds first-class trust-root support for GitHub artifact attestations while keeping instance selection explicit.

GitHub artifact attestations use GitHub's own Sigstore instance, not the public-good Sigstore root. This PR adds the GitHub TUF URL/root constants and embedded GitHub `trusted_root.json` snapshot so callers can actively choose the GitHub instance when verifying GitHub attestations.

The preferred API remains TUF-backed loading via `TrustedRoot::from_tuf(...)`. For GitHub, `TufConfig::github()` is available, but live fetching is not wired as the default because the current `tough` verifier rejects GitHub's TUF root metadata key IDs. Until that compatibility issue is fixed, GitHub verification can explicitly opt into the embedded GitHub trusted-root snapshot with `TrustedRoot::from_embedded(SigstoreInstance::GitHub)`.

This also keeps verification policy explicit: GitHub artifact attestations should be verified with the GitHub root and a policy that skips Rekor/tlog and SCT checks while preserving certificate-chain and RFC 3161 timestamp verification.

#### Release Note

Added GitHub artifact attestation trust-root support, including GitHub TUF configuration and an explicit embedded GitHub trusted-root fallback.

#### Documentation

Updated crate READMEs with explicit public Sigstore/GitHub trust-root selection and GitHub verification usage.


jku:
I can have a closer look on mon/tue but as initial comment:
* I'm not sure TrustRootSet is a good idea? I agree selecting an instance should be easy and we should carry TUF urls for GitHub (or I guess embedded trust root until TUF works) to enable that ease-of-use... but should anyone ever really trust multiple instances when verifying a signature? This feels like a decision that should be made actively, and not be hidden on some list of instances that might change release by release
* This also seems to push more use of embedded sigstore.dev trusted root (through TrustRootSet): we really should be pushing users to TUF, not the embedded trustroot

Adding the Github embedded trustroot without TUF support is annoying but understandable if there's yet another compat issue with the TUF repos, so otherwise this seems good to me.

jku (approved):
Thanks, this looks reasonable to me, no notes.

Issue #89, opened 2026-05-20
Replace chrono with jiff?


ww-oai:
**Description**

I'm looking at integrating `sigstore-rust` into uv, but several of the support crates require chrono. This isn't an issue per se, but it _does_ make our builds much larger (and brings in a transitive closure of deps that we otherwise don't depend on).

An alternative to this is to replace `chrono` with `jiff`, which is (1) widely adopted in the Rust ecosystem, (2) smaller, and (3) generally exposes more ergonomic/misuse-resistant APIs. The downside to this is that it'll be a breaking change, since some of the `sigstore-*` support crate public APIs currently expose `chrono` APIs directly. However, given that this is pre-1.0, I think this is an acceptable breakage for a minor release.

(Separately, I believe this replacement would benefit other downstreams, e.g. the pixi ecosystem that pulls in uv's support crates and therefore could also dedupe time handling code, since uv already uses jiff.)

ww-oai:
I've opened #90 to demonstrate this.

wolfv:
Thanks! I'm fine with moving to jiff - will take a look at the PR today.

PR #90, opened 2026-05-20
Replace direct chrono usage with jiff

ww-oai:
#### Summary

This replaces _direct_ usages of `chrono` with `jiff`. Indirect usages are still present via `tough`; I've filed https://github.com/awslabs/tough/issues/942 for that.

See #89 for motivation.

#### Release Note

This will be a breaking change, since the following public APIs previously exposed chrono interfaces (and with this change expose jiff ones instead):

```
sigstore_tsa::TimestampResult::time
sigstore_trust_root::ServiceValidityPeriod
TrustedRoot::tsa_validity_for_time
TrustedRoot::is_timestamp_within_tsa_validity
TsaCertWithValidity
```

#### Documentation

Only generated documentation changes with this.

### DIFF
--- a/crates/sigstore-tsa/src/verify.rs
+++ b/crates/sigstore-tsa/src/verify.rs
@@ -8,10 +8,10 @@
 
 use crate::asn1::{self, PkiStatus, TimeStampResp, TstInfo};
 use crate::error::{Error, Result};
-use jiff::Timestamp;
 use cms::cert::CertificateChoices;
 use cms::signed_data::{SignedData, SignerIdentifier};
 use const_oid::ObjectIdentifier;
+use jiff::Timestamp;
 use rustls_pki_types::CertificateDer;
 use x509_cert::Certificate;
 
--- a/crates/sigstore-tsa/src/verify.rs
+++ b/crates/sigstore-tsa/src/verify.rs
@@ -8,7 +8,7 @@
 
 use crate::asn1::{self, PkiStatus, TimeStampResp, TstInfo};
 use crate::error::{Error, Result};
-use chrono::{DateTime, Utc};
+use jiff::Timestamp;
 use cms::cert::CertificateChoices;
 use cms::signed_data::{SignedData, SignerIdentifier};
 use const_oid::ObjectIdentifier;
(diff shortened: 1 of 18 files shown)


wolfv (dismissed):

wolfv:
Ah darn, rustfmt :) Happy to merge this PR. Would be ideal if `tough` also switches. It's a refactor we've planned for rattler / Pixi as well.

ww-oai:
Oops, I'll fix that. Yeah, I'll coordinate a similar switch with tough!

woodruffw (approved):

woodruffw:
(Stamping my own changes since the only difference since last approval is `cargo fmt`.)

crates/sigstore-tsa/src/verify.rs
```diff
@@ -204,9 +204,10 @@ pub fn verify_timestamp_response(
     let unix_duration = system_time
         .duration_since(std::time::UNIX_EPOCH)
         .map_err(|_| Error::ParseError("timestamp before epoch".to_string()))?;
-    let timestamp =
-        DateTime::from_timestamp(unix_duration.as_secs() as i64, unix_duration.subsec_nanos())
-            .ok_or_else(|| Error::ParseError("invalid timestamp in TSTInfo".to_string()))?;
+    let seconds = i64::try_from(unix_duration.as_secs())
+        .map_err(|_| Error::ParseError("invalid timestamp in TSTInfo".to_string()))?;
+    let timestamp = Timestamp::new(seconds, unix_duration.subsec_nanos() as i32)
+        .map_err(|_| Error::ParseError("invalid timestamp in TSTInfo".to_string()))?;
```
BurntSushi, line 210: I'd go this route personally, since Jiff's `TryFrom<SystemTime> for Timestamp` impl handles all this fiddliness:

```suggestion
    let timestamp = Timestamp::try_from(tst_info.gen_time.to_system_time())
        .map_err(|_| Error::ParseError("invalid timestamp in TSTInfo".to_string()))?;
```

This will actually accept more inputs than previously, since this handles the case of a system time being negative. It's an oddball case, but if you want this to be exact, then I'd do something like this:

```suggestion
    let timestamp = match Timestamp::try_from(sys) {
        Ok(timestamp) if timestamp >= Timestamp::UNIX_EPOCH => timestamp,
        _ => return Err(Error::ParseError("invalid timestamp in TSTInfo".to_string())),
    };    
```

woodruffw: Thanks! I'll do a follow-up PR for that.

PR #97, opened 2026-05-21
Make aws-lc-rs optional when using native-tls feature

r0x0d:
#### Summary

Previously, aws-lc-rs was unconditionally pulled in through three paths even when the native-tls feature was selected:

1. sigstore-crypto depended on aws-lc-rs unconditionally
2. sigstore-tsa had a direct aws-lc-rs dependency
3. rustls-webpki workspace config hardcoded the aws-lc-rs feature

This change introduces ring as an alternative crypto backend, feature-gated behind rustls/native-tls flags:

- sigstore-crypto: Add rustls/native-tls features where rustls maps to aws-lc-rs and native-tls maps to ring. All source files use cfg to select the backend (the APIs are nearly identical).
- sigstore-tsa: Remove direct aws-lc-rs dep, add optional aws-lc-rs/ring gated on features, cfg-gate verify.rs digest and signature usage.
- sigstore-verify: Feature-gate rustls-webpki backend (aws-lc-rs vs ring).
- All dependent crates: Propagate sigstore-crypto features through their existing rustls/native-tls feature gates.
- check-native-tls.sh: Now checks for both rustls and aws-lc-rs as forbidden crates.

Verified with:
- cargo check (default features): full workspace builds successfully
- cargo check -p sigstore-sign --no-default-features --features native-tls: builds without aws-lc-rs
- cargo tree -i aws-lc-rs -p sigstore-sign --no-default-features --features native-tls: "package ID specification `aws-lc-rs` did not match any packages"
- cargo tree -i aws-lc-rs -p sigstore-verify --no-default-features --features native-tls: "package ID specification `aws-lc-rs` did not match any packages"
- cargo test: all 400+ tests pass with default features
- bash scripts/check-native-tls.sh: 12 checked, 0 failed, 1 skipped

#### Release Note

* Make aws-lc-rs optional: selecting the native-tls feature now uses ring as the crypto backend, fully eliminating the aws-lc-rs dependency for environments where it is unwanted or difficult to build.

#### Documentation

No documentation change required.


r0x0d:
Hi! I'm not sure if this is the type of thing that I need to create an issue for, but, in case it is, I can definitely do that. 

Just to add a bit of context to this PR, I'm taking care of packaging/updating https://github.com/aaif-goose/goose in Fedora/EPEL/RHEL, and currently, to [move goose to the latest version](https://github.com/rhel-lightspeed/goose/pull/27), we would require native-tls to be supported fully and drop the aws-lc-* references. 

Since I'm a new contributor here, please, let me know what is the correct process for making such change like that here.

wolfv:
I think we can close this. The `reqwest` story should be quite clean, and aws-lc-rs is used for the crypto part. If you want to make the crypto backend exchangeable for ring or something else, maybe you can start a new PR on top of current main!

PR #99, opened 2026-05-26
Support dsse as hashedrekord

jku:
### Rekor2: Support DSSE via Hashedrekord entries
    
 This is based on https://github.com/sigstore/architecture-docs/pull/63
 * With Rekor v2 DSSE envelopes are supported via Hashedrekord entries instead of a dedicated dsse entry
* digest is Hash(PAE(payloadType, payload))
* Signature is the DSSE envelope signature
    
This removes all support for the "dsse 0.0.2" entry type: this seems ok as rekor v2 signing was never deployed in production (the log is there, it's just never been included in the SigningConfig).
* When signing with rekor v2, build a "hashedrekord 0.0.2" entry with digest and signature as described above
* When verifying hashedrekord, compute hash depending on bundle content
* When matching hashedrekord signatures, use envelope signatures when appropriate
* When verifying rekor entry consistency, accept a DSSE envelope with a "hashedrekord 0.0.2" entry
    
Testing:
* Implement a few new verification tests
* Use conformance main for now (tests for this are not released yet)

The test asset is from 
`cargo run -p sigstore-sign --example sign_attestation -- --staging crates/sigstore-verify/test_data/bundles/signed-package-2.1.0-hb0f4dca_0.conda -o crates/sigstore-verify/test_data/bundles/conda-attestation-rekor2.sigstore.json
`
and verifies with:
`cargo run -p sigstore-verify --example verify_bundle --     --staging     crates/sigstore-verify/test_data/bundles/signed-package-2.1.0-hb0f4dca_0.conda     crates/sigstore-verify/test_data/bundles/conda-attestation-rekor2.sigstore.json`


jku:
I've opened #102, #103 as better defined, more focused PRs from this one: once those have been somehow handled I will update this branch so it only includes the dsse-as-hashedrekord change

jku:
Currently using a commit from conformance main to get the new tests for the dsse-as-hashedrekord: next conformance release will contain them

jku:
rebased on both #102 and #103.

jku:
I suppose this is ready for review

crates/sigstore-sign/src/sign.rs
```diff
@@ -456,25 +454,33 @@ impl Signer {
...
                     Error::Signing(format!("Failed to create DSSE Rekor entry: {}", e))
                 })?;
-                (entry, "0.0.1")
+                (entry, "dsse", "0.0.1")
             }
             RekorApiVersion::V2 => {
-                let dsse_entry = DsseEntryV2::new(envelope, certificate);
-                let entry = rekor.create_dsse_entry_v2(dsse_entry).await.map_err(|e| {
-                    Error::Signing(format!("Failed to create DSSE Rekor entry: {}", e))
+                let hash = sigstore_crypto::sha256(&envelope.pae());
```
loosebazooka, line 473: is there no standard rust crypto that does sha256?

crates/sigstore-sign/src/sign.rs
```diff
@@ -456,25 +454,33 @@ impl Signer {
...
+
+                // Extract the first signature
```
loosebazooka: should the client enforce 1 sig per dsse? as per the sigstore spec?

crates/sigstore-sign/src/sign.rs
```diff
@@ -456,25 +454,33 @@ impl Signer {
...
+                let signature =
+                    envelope.signatures.first().map(|s| &s.sig).ok_or_else(|| {
+                        Error::Signing("DSSE envelope has no signatures".to_string())
+                    })?;
+
+                let hashed_rekord = HashedRekordV2::new(&hash, signature, certificate);
```
loosebazooka, line 477: I'm a little out of touch with the rust memory model, but if `hash` isn't used again, can you just assign ownership to `HashedRekordV2::new` ? Maybe this doesn't matter and I'm just confused

crates/sigstore-sign/src/sign.rs
```diff
@@ -456,25 +454,33 @@ impl Signer {
         let rekor = RekorClient::new(&self.rekor_url);
 
         // Use V1 or V2 API based on configuration
-        let (log_entry, version) = match self.rekor_api_version {
+        let (log_entry, kind, version) = match self.rekor_api_version {
```
loosebazooka, line 464: how is `log_entry` typed here (just to the closest viable parent?)

crates/sigstore-verify/src/verify_impl/hashedrekord.rs
```diff
@@ -27,24 +27,27 @@ pub(crate) fn verify_hashedrekord_entry(
     )
     .map_err(|e| Error::Verification(format!("failed to parse Rekor body: {}", e)))?;
 
-    // Compute artifact hash from artifact (bytes or pre-computed digest)
-    let artifact_hash = compute_artifact_digest(artifact);
+    // Compute hash from artifact (bytes or pre-computed digest) or DSSE envelope
+    let hash = match &bundle.content {
+        SignatureContent::MessageSignature(_) => compute_artifact_digest(artifact),
+        SignatureContent::DsseEnvelope(envelope) => sigstore_crypto::sha256(&envelope.pae()),
```
loosebazooka, line 33: I thought you needed to use the algorithm_registry to get the signing algorithm to determine the hash algorithm. (maybe rust is just sha256 for now?)

crates/sigstore-verify/src/verify_impl/hashedrekord.rs
```diff
@@ -27,24 +27,27 @@ pub(crate) fn verify_hashedrekord_entry(
...
+    };
 
     // Validate artifact hash matches what's in Rekor
     match &body {
         RekorEntryBody::HashedRekordV001(rekord) => {
             // v0.0.1: spec.data.hash.value (hex-encoded)
             let expected = Sha256Hash::from_hex(rekord.spec.data.hash.value.as_str())
                 .map_err(|e| Error::Verification(format!("invalid hash in Rekor entry: {}", e)))?;
-            validate_artifact_hash(&artifact_hash, &expected)?;
+            validate_artifact_hash(&hash, &expected)?;
```
loosebazooka, line 42: this logic seems to allow the dsse as hashedrekord in rekor v1 as well? Is that what we want? (it seems like there is no consequence to it)

loosebazooka (commented):
mostly questions about rust for me to understand.

crates/sigstore-sign/src/sign.rs
jku: not without changing the HashedRekordV2::new function signature. 

I *think* a change like that would actually be pretty idiomatic (since hash is basically a `[u8]` and simple things like that usually are passed by value) but in practice this doesn't seem important.

crates/sigstore-sign/src/sign.rs
jku: log_entry is always a  `LogEntry` (the rekor v2 code path has some "conversion code" that I'm not yet familiar with)

crates/sigstore-verify/src/verify_impl/hashedrekord.rs
jku: here it does not error but `verify_tlog_consistency()` only accepts specific combinations: there is even a test case for that

crates/sigstore-sign/src/sign.rs
jku: There indeed is no standard crypto in Rust. 

We use aws-lc IIRC. I don't know why we use our own wrapper but my guess would be:
* hide aws-lc so it can be swapped out if needed
* get some nicer de/serialization (to base64 etc)

crates/sigstore-verify/src/verify_impl/hashedrekord.rs
jku: Yeah, I think you are correct: there is no dynamic hash selection for hashedrekord right now. Filed #104 for this

We should have a test for this in conformance -- cover the signing algorithms we actually care about -- otherwise this is not going to actually work in any client

wolfv (approved):

PR #103, opened 2026-05-27
verify: Refactor log entry consistency verification

jku:
Verify that
1. We understand each entry (kind and version)
2. Each entry matches the bundle material

The old code seemingly only verified the second part if we happened to recognise the entry type (unknown types and versions were just skipped)

Additionally, only runs this code if policy includes "verify_tlog".

### DIFF
--- a/crates/sigstore-verify/src/verify.rs
+++ b/crates/sigstore-verify/src/verify.rs
@@ -415,9 +415,9 @@ impl Verifier {
 
         // (8): Verify the transparency log entry's consistency against the other
         //      materials, to prevent variants of CVE-2022-36056.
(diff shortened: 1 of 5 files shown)


jku:
I'm not really 100% sure about this design but it seems like an improvement:
* maybe it would be useful to expose consistency verification for specific types individually in the API... 
* but it seems just as possible that we could just extend the policy mechanism (so users could define what types they expect to find) -- and and in reality in the future (some day when we're rekor v2 only) we only expect hashedrekord
* in any case, I think the default case should be what this PR does: we have a list of known entry types we can verify, all entries should be in that list, and we should verify all entries

wolfv (approved):

PR #106, opened 2026-06-01
Add sigstore-tuf crate; bootstrap trusted roots over TUF without tough

wolfv:
## Summary

Introduces **`sigstore-tuf`**, a pure-Rust [TUF](https://theupdateframework.io/) client built for Sigstore and interoperable with the whole ecosystem — including `tuf-on-ci` / `securesystemslib`-produced roots such as **GitHub's** (`https://tuf-repo.github.com`), which `tough` cannot load — and rewires `sigstore-trust-root` onto it, dropping the `tough` dependency.

### Why a new crate?

`tough` recomputes each key's TUF key ID over the entire key object (including non-standard fields like `x-tuf-on-ci-keyowner`) and *rejects* metadata whose declared key IDs don't match. That makes it unable to bootstrap GitHub's TUF root at all, even though `python-tuf`, `go-tuf`, and `gh attestation` consume it fine. Per the TUF spec a key ID is an opaque, producer-chosen identifier.

Two interop decisions are central:

- **Declared key IDs are authoritative** (matching `python-tuf` / `go-tuf`). We never recompute-and-reject — this is what makes GitHub's root load.
- **Canonical JSON is securesystemslib (OLPC), not RFC 8785 JCS.** The two differ in string escaping (notably newlines inside embedded PEM keys) and number handling, so a JCS encoder would fail every signature check. The crate ships its own encoder.

## What's in `sigstore-tuf`

- Metadata model + signed envelope (preserves the canonical signed bytes the signatures actually cover)
- securesystemslib canonical JSON
- Key handling → `sigstore-crypto` verification keys (ecdsa/rsa PEM SPKI, ed25519 raw-hex)
- `TrustedMetadataSet` — transport-free verification state machine: signature thresholds, anti-rollback, expiry, length/hash pinning
- Pluggable `Repository` transport with HTTP + offline/cache implementations
- Per-file download size limits (`UpdaterConfig`, python-tuf defaults) with memory bounded while streaming
- Write-through on-disk caching that re-verifies the root chain from the pinned bootstrap root
- Delegated-targets discovery (pre-order DFS, `terminating`, `paths` globs + `path_hash_prefixes`), verified against the delegator's keys

## Wiring

`sigstore-trust-root`'s TUF path now uses `sigstore_tuf::Updater` instead of `tough::RepositoryLoader`; `tough` (and now-unused `futures`/`url`/`reqwest_0_12`) are removed. TLS follows the workspace `rustls`/`native-tls` feature idiom. **Public API is unchanged** (`TufConfig`, `from_tuf`, `production`/`staging`/`github`, `offline`, caching, `fetch_trust_material`).

## Testing

- Unit: canonical JSON, glob matching
- Loads the real embedded GitHub + Sigstore production/staging roots and self-verifies them (the GitHub case fails under `tough`); tampering breaks verification
- A fully-signed in-memory repo (ECDSA P-256) exercising root → timestamp → snapshot → **delegation hand-off to a different key** → target download → offline cache re-verify
- Network-gated (`#[ignore]`) live bootstrap of production **and** GitHub trusted roots — both pass against the live repos

`cargo clippy --workspace --all-targets` is clean; full workspace test suite passes; `tough` is gone from the dependency tree.


crates/sigstore-tuf/src/metadata/role.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// Whether this delegated role is authorized for `target_path`.
+    ///
+    /// A role authorizes a path if it matches one of its `path_hash_prefixes`
+    /// (the SHA-256 hex of the target path begins with the prefix) or one of its
+    /// shell-style `paths` globs. A role with neither field authorizes nothing.
+    ///
+    /// Glob matching uses [`globset`] with its default options, which match
+    /// `python-tuf`'s `fnmatch` semantics (a `*` *does* span `/`) — the same
+    /// crate and configuration `tough` uses. A `paths` entry that is not a valid
```
jku: I have a feeling this is incorrect: tuf glob semantics are basically _shell glob_ so "*/c" should not match "a/b/c", it should only match "a/c".

the relevant tests look incorrect to me as well -- I might be wrong but I expect conformance to complain about this

wolfv: I will double check that against the implementations in Go and Python. Do you know which one is the best implementation to take as a reference?

jku: I'm happy with python-tuf but then current go implementation is heavily based on python-tuf so :shrug:

jku: I looked and I'm quite sure I am right -- the spec even has an example of this. Added tests in https://github.com/theupdateframework/tuf-conformance/pull/388

jku:
there's a minor issue with offline mode and the way metadata is cached but I don't know if this worth solving
* a delegated role could use a role name like "timestamp" or "2.root"  
* if the server was compromised and actually returned a targets metadata when the url is requested, the targets metadata would likely get accepted by client and would overwrite the cached top-level role
* when the client next starts up, the top-level role in the cache would fail to verify and client would have to download it again
* if the client was offline on this second run, this would be a DOS

It does not seem like there is a real attack here but I'm not 100%. Some potential improvements that wouldn't require massive redesign:
* we could prevent  ("root", "timestamp", "snapshot") from being used as delegated targets role name just in case
* could cache the versioned root files in another namespace to prevent any funny business (python-tuf uses `<metadata-cache>/root_history/1.root.json`)

crates/sigstore-tuf/Cargo.toml
```diff
@@ -0,0 +1,57 @@
...
+url = { workspace = true, optional = true }
+reqwest = { workspace = true, optional = true }
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
+sigstore-crypto = { workspace = true }
+serde_json = { workspace = true }
+sha2 = { workspace = true }
+hex = { workspace = true }
+jiff = { workspace = true }
```
jku: do we need to repeat these in dev-dependencies?

crates/sigstore-verify/README.md
```diff
@@ -60,10 +60,9 @@ use sigstore_types::{Bundle, Sha256Hash};
 let bundle: Bundle = serde_json::from_str(bundle_json)?;
 let artifact_digest = Sha256Hash::from_hex("...")?;
 
-// Preferred API shape once GitHub TUF is compatible with the TUF client:
+// Fetch GitHub's trusted root over TUF (now supported via the `sigstore-tuf`
+// client), or use the embedded copy below for an offline path.
 // let root = TrustedRoot::from_tuf(sigstore_trust_root::TufConfig::github()).await?;
-
-// Temporary explicit fallback while GitHub TUF metadata is not accepted by `tough`.
 let root = TrustedRoot::from_embedded(SigstoreInstance::GitHub)?;
```
jku, line 66:
```suggestion
// Fetch GitHub's trusted root over TUF
let root = TrustedRoot::from_tuf(sigstore_trust_root::TufConfig::github()).await?;
```

crates/sigstore-tuf/src/client.rs
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    /// Download a target file and verify its length and hash against the trusted
+    /// targets metadata (searching delegations). Requires a prior
+    /// [`Updater::refresh`].
+    pub async fn download_target(
+        &mut self,
+        target_path: &str,
+        now: jiff::Timestamp,
+    ) -> Result<Vec<u8>> {
```
jku: two notes on this method:
* `target_path: &str` as argument makes the simple example case simple... but makes real usage less efficient: caller first needs to call `get_targetinfo()` so they can check if they actually need to download or not, then they call `download_target()` but that always calls `get_targetinfo()` again. This is why in python-tuf download_target() takes a TargetInfo as argument 
* A generic tuf implementation would likely want to provide a streaming solution as well since the target could be gigabytes -- but for sigstore this is fine and the streaming is a bit tricky since we do still want to verify the hash

crates/sigstore-trust-root/src/tuf.rs
```diff
@@ -321,8 +320,11 @@ impl TufClient {
         if self.config.offline {
             return self.fetch_target_offline(target_name).await;
         }
-        let repo = self.load_repository().await?;
-        self.read_target_from_repo(&repo, target_name).await
+        let mut updater = self.build_updater().await?;
+        updater
+            .download_target(target_name, jiff::Timestamp::now())
+            .await
+            .map_err(|e| Error::Tuf(format!("Failed to fetch target {target_name}: {e}")))
```
jku, line 327: so there is a workaround in conformance to pass the caching test... but our actual implementation seems to download the target file every time it runs?

jku (changes requested):
I've read through this and I think it is a fine TUF client. It's quite a lot of code to maintain but I guess we can do it.

The only significant ask I have is to review the target (trustedroot and signingconfig) caching:
* The TUF client leaves local file IO  to caller (for targets). This is fine except ... 
* it looks like we don't actually look for the trust root files in a cache -- the files are stored in the cache but we always download them again
* the conformance test client does implement caching

There is a related comment about the API -- the way get_targetinfo and download_target are defined is a bit inefficient and leads to user mistakes like the above cache miss.

crates/sigstore-trust-root/src/tuf.rs
wolfv: yep, should be fixed with the last commit

jku (approved):
seems good.

* non-atomic writes to disk could always be an issue -- but the way you handle the cache means this should never be that harmful
* the lack of streaming API now applies to all of get_target/find_cached_target/download_target... but for our use case this is fine
* we may want to spend some time considering concurrency at some point (this code likely would have issues if multiple instances were running on windows) but it feels fine for now

I guess we can try this?

PR #109, opened 2026-06-03
crypto: Various improvements

jku:
This ended up becoming multiple changes:

* Avoid double-hashing digests via verify_inner for non-ECDSA schemes
* Do not fail-open when unsupported signing schemes are seen
* Support multiple digest types via a "DigestBytes" and generic slices: in practice this means sha256/sha384/sha512. The hash algorithm is chosen based on the signing algorithm, except...
* Introduces KeyAlgorithm enum in to represent public key types extracted from the certificate without guessing the hash algorithm: in some cases we can get the actual hash algorithm from the bundle

### DIFF
--- a/crates/sigstore-crypto/src/hash.rs
+++ b/crates/sigstore-crypto/src/hash.rs
@@ -1,6 +1,6 @@
 //! Hashing utilities using aws-lc-rs
 
-use aws_lc_rs::digest::{self, Context, SHA256};
+use aws_lc_rs::digest::{self, Context, SHA256, SHA384, SHA512};
 use sigstore_types::Sha256Hash;
 use std::io::{self, Read};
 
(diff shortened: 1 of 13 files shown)


crates/sigstore-crypto/src/hash.rs
```diff
@@ -12,6 +12,18 @@ pub fn sha256(data: &[u8]) -> Sha256Hash {
...
+pub fn sha384(data: &[u8]) -> Vec<u8> {
+    let digest = digest::digest(&SHA384, data);
+    digest.as_ref().to_vec()
+}
+
+/// Hash data using SHA-512, returning raw bytes
+pub fn sha512(data: &[u8]) -> Vec<u8> {
+    let digest = digest::digest(&SHA512, data);
+    digest.as_ref().to_vec()
+}
```
jku, line 25: these are currently *not* quite like the sha256() that returns a Sha256Hash... that one is used extensively so feels bad to modify

jku:
> * Introduces KeyAlgorithm enum in to represent public key types extracted from the certificate without guessing the hash algorithm

this happened because some of the conformance tests started failing now that the code is more strict -- there are cpython release bundles with an unusual combination of hash & signing algorithm. We could just mark those xfailed, I don't think the bundles are correct... but I did make it work: as long as the bundle contains the digest algorithm, we can figure it out for even these cases.

jku:
The external changes should be here but this did become much more invasive than I hoped: @wolfv if you have suggestions, I can rework this.

### sigstore-crypto
* Replaced `CertificateInfo::signing_scheme` with `key_algorithm`: allows key algorithms to dynamically resolve to a  SigningScheme based on bundle content
* Expanded SigningSchemes to cover a few used in conformance
* Generalized prehashing to accept &[u8] digests

### sigstore-types
* exposed DigestBytes as generic digest type, made sure it converts nicely with Sha256Hash
* safer artifact construction without accidental double-hashing

### sigstore-verify
* verification dynamically resolves correct digest algorithm from the bundle
* Avoid failing open when bundle contains unknown signingschemes

wolfv (approved):

PR #111, opened 2026-06-08
refactor: simplify conformance client using native DSSE validation

jku:
It's possible I misunderstood something but it looks like conformance client has a redundant manual check:
* Remove seemingly redundant manual DSSE payload parsing in conformance -- this is done in sigstore-verify
* Add test to verify DSSE artifact matching


wolfv (approved):
Thanks!

PR #116, opened 2026-06-10
Merkle fixes

jku:
Two separate changes that I can split if requested:
* Fix overflow in max tree size case
* Remove helper methods that do basically nothing (stdlib `trailing_zeros()` and `leading_zeros()` already handle the case of 0 as input correctly, no need to special case this locally)


wolfv (approved):
Thanks!

PR #117, opened 2026-06-10
sign: Require complete SigningConfig from tuf

jku:
Let's not use default values if TUF is used and the received SigningConfig is incomplete


crates/sigstore-sign/src/sign.rs
```diff
@@ -80,7 +82,7 @@ impl SigningConfig {
     /// # Arguments
     ///
     /// * `tuf_config` - The signing config from TUF
-    pub fn from_tuf_config(tuf_config: &TufSigningConfig) -> Self {
+    pub fn from_tuf_config(tuf_config: &TufSigningConfig) -> Result<Self> {
```
jku, line 85: if there is a good way to do this without the API change I could try that too

wolfv (approved):

PR #119, opened 2026-06-10
verify: harden message-signature verification, drop digest leak

### LINKED ISSUE #115: fail-open verify for empty dsse subject lines
we do this in DSSE verification (two separate places)

```
if !statement.subject.is_empty() && !statement.matches_sha256(&artifact_hash_hex) {
    return Err(Error::Verification(
        "artifact hash does not match any subject in attestation".to_string(),
    ));
}
```

empty subject string does not seem like a useful thing to special case

wolfv:
Follow-up cleanups on top of #109.

## Changes

**Drop the `Box::leak` digest leak.** `From<Sha256Hash> for Artifact<'static>` previously leaked 32 bytes per conversion via `Box::leak`. `Artifact::Digest` now holds a `Cow<'a, [u8]>`: borrowed conversions stay zero-copy, owned conversions copy instead of leaking. Adds a matching owned `From<DigestBytes>`.

**Always verify the MessageSignature signature.** In `Verifier::verify`, cryptographic signature verification for `MessageSignature` bundles previously only happened inside the transparency-log path (step 8), gated on `policy.verify_tlog`. With tlog verification disabled, only the `messageDigest`-vs-artifact comparison ran and the signature itself was never checked. It now runs unconditionally.

> Note: this is a behavioral change — bundles that previously "verified" with `verify_tlog = false` purely on a digest match now require a valid signature.

**Deduplicate signature-verification logic.** The "resolve scheme → branch on bytes/digest → fail closed if unsupported" logic was copied across `hashedrekord`, `verify_with_key`, and the new path. Consolidated into a single `verify_signature_over_artifact` core plus a cert-based `verify_message_signature_crypto` wrapper. Removes the now-unused `compute_artifact_digest_for_scheme` (~135 lines of duplication removed).

**Test coverage.** Adds a P-384/SHA-384 round-trip test confirming the standard pairing verifies (raw + prehashed) and that a mismatched SHA-256 scheme resolution fails closed rather than silently accepting.

## Testing

`cargo test` (crypto/types/verify), `cargo clippy --all-targets`, and `cargo fmt` all clean.

### DIFF
--- a/crates/sigstore-verify/src/verify_impl/hashedrekord.rs
+++ b/crates/sigstore-verify/src/verify_impl/hashedrekord.rs
@@ -10,8 +10,14 @@ use sigstore_types::{Artifact, Bundle, Sha256Hash, SignatureContent, Transparenc
 use x509_cert::der::Decode;
 use x509_cert::Certificate;
 
-/// Verify artifact/dsse hash matches what's in Rekor (for hashedrekord entries)
-/// Verify a single hashedrekord entry
+/// Verify a single hashedrekord entry's consistency against the bundle and
+/// artifact: the artifact/DSSE hash, certificate, and signature recorded in
+/// Rekor must match the bundle's materials.
+///
+/// This does NOT cryptographically verify the signature itself; that happens
+/// unconditionally in `Verifier::verify` (step 7). Since the Rekor entry's
+/// signature is checked for equality with the bundle's signature here, that
+/// single verification covers both.
 pub(crate) fn verify_hashedrekord_entry(
     entry: &TransparencyLogEntry,
     bundle: &Bundle,
@@ -63,9 +69,6 @@ pub(crate) fn verify_hashedrekord_entry(
 
     // Validate integrated time is within certificate validity (for v0.0.1)
     validate_integrated_time(entry, bundle)?;
-
-    // Perform cryptographic signature verification
-    verify_signature_cryptographically(bundle, artifact)?;
 
     Ok(())
 }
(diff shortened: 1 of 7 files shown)


crates/sigstore-verify/src/verify_impl/hashedrekord.rs
```diff
@@ -67,7 +65,7 @@ pub(crate) fn verify_hashedrekord_entry(
     validate_integrated_time(entry, bundle)?;
 
     // Perform cryptographic signature verification
-    verify_signature_cryptographically(entry, &body, bundle, artifact)?;
+    verify_signature_cryptographically(bundle, artifact)?;
```
jku: Do we even want to do this here anymore -- I think step 7 in Verifier.verify() now handles the signature check consistently already? 

We could potentially remove `verify_signature_cryptographically()` altogether

jku (dismissed):
oh yeah, much better. This looks good to me but I left one further suggestion: How about we don't do signature checks in the hashedrekord verify at all because we now already do them in step 7 -- maybe the hashedrekord docstring needs to mention that signature is not verified here but this feels like ok design: it removes another case of what looks like fail-open verify (it isn't really that in the end but _"Only verify for MessageSignature"_ always looks fishy in a comment).

let me know if I misunderstood that.

crates/sigstore-verify/src/verify_impl/hashedrekord.rs
jku, line 143: another fail-open here -- this one seems correct at the moment but covering all of the RekorEntryBodies we expect would likely be more resilient

jku (approved):
Nice, lgtm!

We really should try to split these into more PRs in future. I'm guilty of bundling many fixes as well, the AI really likes to do it and sometimes the changes are related... but reviewing and iterating would be much easier if unrelated changes were separate PRs

wolfv:
Yes, will keep that in mind for the future!

PR #120, opened 2026-06-12
ci: drop dtolnay/rust-toolchain in favor of rustup + rust-toolchain.toml

### LINKED ISSUE #95: toolchain pinning
A recent PR Added this:
```
      - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # v1
        with:
          toolchain: 1.89.0
```

Pinning anything like this has a bit of a smell: it will probably not be updated again for a long time.

I'm not very experienced with rust toolchain management but is there an advantage to using  dtolnay/rust-toolchain versus just running rustup (which is in the runner images) and maintaining a rust-toolchain.toml file (that dependabot IIUC can work with)?

CC @tnytown

wolfv:
Per the discussion in #95: the `dtolnay/rust-toolchain` action mostly shells out to rustup and we don't use its matrix-parameterization capability, so this drops it entirely.

- Adds `rust-toolchain.toml` at the repo root pinning `1.89.0` with the `clippy` and `rustfmt` components (one canonical definition, also used by local dev).
- Replaces all 10 action usages across `ci.yml`, `interop.yml`, and `release-plz.yml` with the same one-liner:
  ```yaml
  - name: Install Rust toolchain
    run: rustup show active-toolchain || rustup toolchain install
  ```
  This is the rustup-documented idiom: a no-op when the toolchain from `rust-toolchain.toml` is already installed, otherwise `rustup toolchain install` (no args) installs the pinned channel plus its components. It works on both pre- and post-1.28 rustup.

Notes:
- No job used `targets:`; only the fmt and clippy jobs carried `components:`, which moved into `rust-toolchain.toml`.
- zizmor v1.25.2 reports no findings on the modified workflows; YAML validated.
- There is currently no dependabot/renovate config in the repo, so nothing auto-updates the pin yet. Dependabot has a `rust-toolchain` ecosystem nowadays — happy to add a `.github/dependabot.yml` as a follow-up if wanted.

Fixes #95


jku (dismissed):
I've not done a lot of rust maintenance but this makes a lot of sense:
* one less gh action to update
* toolchain config in a single, expected place

Enabling dependabot sounds good, here or in different PR WFM

jku (approved):

PR #122, opened 2026-06-12
trust-root: enforce valid_for in key accessors, surface parse errors

### LINKED ISSUE #112: trust-root: some convenience methods disregard valid_for
It looks like some valid_for constraints are not checked when returning values in methods like rekor_keys()

Also, valid_for parsing seems to currently swallow errors silently, that seems incorrect.

wolfv:
Fixes the two problems from #112: accessors like `rekor_keys()` ignored `valid_for`, and `valid_for` timestamp parsing silently swallowed errors (`.parse().ok()` treated malformed timestamps as unbounded validity).

**Semantics chosen** (matching sigstore-go's `RekorLogs` and sigstore-python's verify-path behavior):

- All verification-material accessors (`rekor_keys`, `rekor_keys_with_hints`, `rekor_key_for_log`, `ctfe_keys`, `ctfe_keys_with_ids`, `fulcio_certs`, `tsa_root_certs`, `tsa_intermediate_certs`, `tsa_leaf_certs`) now **exclude instances whose `valid_for.start` is in the future** and **propagate timestamp parse errors** instead of ignoring them.
- **Expired instances are deliberately kept**: they're required to verify historical material (the embedded production/GitHub trusted roots contain expired ctlogs/CAs/TSAs that are still needed to verify old bundles). Filtering to a strict now-window would have broken verification of all pre-2025 GitHub attestations.
- Where the relevant time IS known, the full window is enforced: new `rekor_key_for_log_at(log_id, time)` requires the window to contain `time`, and `verify_set()` now uses it with the entry's integrated time — a SET only verifies if the entry was integrated while the log key was valid (matches sigstore-go).
- New `ValidityPeriod` API: `start()`, `end()`, `contains(time)`, `has_started_by(time)`, all returning `Result` so malformed timestamps error.

**Breaking change**: `is_timestamp_within_tsa_validity()` now returns `Result<bool>`; its caller in sigstore-verify fails verification on malformed trust-root timestamps instead of silently passing.

Seven new unit tests cover future-key exclusion, expired-key retention, full-window checks in `rekor_key_for_log_at`, and malformed-timestamp errors across rekor/ctlog/fulcio/TSA accessors. `cargo clippy --workspace -D warnings` clean, full workspace test suite passes.

Fixes #112

### DIFF
--- a/crates/sigstore-trust-root/src/trusted_root.rs
+++ b/crates/sigstore-trust-root/src/trusted_root.rs
@@ -171,6 +171,65 @@ pub struct ValidityPeriod {
     pub end: Option<String>,
 }
 
+fn parse_validity_timestamp(value: Option<&str>, field: &str) -> Result<Option<Timestamp>> {
+    value
+        .map(|s| {
+            s.parse::<Timestamp>().map_err(|e| {
+                Error::TimeParse(format!("invalid validFor.{field} timestamp {s:?}: {e}"))
+            })
+        })
+        .transpose()
+}
+
+impl ValidityPeriod {
+    /// Parsed start of the validity window.
+    ///
+    /// Returns an error if the timestamp is present but malformed.
+    pub fn start(&self) -> Result<Option<Timestamp>> {
+        parse_validity_timestamp(self.start.as_deref(), "start")
+    }
+
+    /// Parsed end of the validity window.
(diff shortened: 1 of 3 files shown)


jku (approved):
Nice, thanks

PR #123, opened 2026-06-12
bundle: separate structural validation from crypto verification

### LINKED ISSUE #110: ambiguous validation in sigstore-bundle
I've been trying to review the verification paths and I think sigstore-bundles `validate_bundle()` is too ambiguous: 
* it's primarily structural validation...
* but it also does crypto validation in `validate_inclusion_proofs()`...
* but then does not verify checkpoint signature...

This makes it hard to understand what the goal of validate_bundle() is. As a consequence verifying that the method does what it intends is also difficult.

Maybe this can be improved with just docstrings, or maybe the API should change?

wolfv:
Resolves the ambiguity described in #110: `validate_bundle()` mixed structural validation with cryptographic Merkle-proof verification, while not verifying checkpoint signatures — making its contract hard to state or audit.

**New contract:**

- `validate_bundle{,_with_options}` is now **purely structural**: media-type/version conformance, version-specific shape rules, required-field presence, and inclusion-proof *well-formedness* (checkpoint parses; `log_index`/`tree_size` valid and `log_index < tree_size`; the proof's `root_hash` matches its own checkpoint's root hash — internal consistency, no keys involved). Docstrings on every public item now explicitly list what is and is **not** checked (no signature, Merkle, checkpoint, SET, timestamp, or certificate verification).
- The cryptographic checks moved to the verify path in sigstore-verify: new `verify_entry_inclusion()` per tlog entry verifies the **Merkle inclusion proof** (leaf = canonicalized body, hashed up to the proof root), the **checkpoint signature** (binding that root hash to a Rekor key from the trusted root), and the **SET**. Both `Verifier::verify` and `verify_with_key` route through it.

**No verification was weakened — coverage is equal or stronger:** previously the Merkle proof was checked against an *unauthenticated* root hash inside `validate_bundle`; now it's always checked in the verify flow against the root hash bound by the signed checkpoint. New contract-pinning tests prove it: a tampered Merkle `root_hash` or canonicalized body now *passes* structural validation but *fails* `verify()`.

Also: `sigstore-merkle` demoted to a dev-dependency of sigstore-bundle, unused `Error::Merkle` variant dropped.

One judgment call worth review: the proof-root-hash vs checkpoint-root-hash consistency check stays in structural validation (it needs no keys) *and* in `verify_checkpoint` (defense in depth). If you'd prefer structural validation to be presence/shape only, it can move out.

`cargo clippy --workspace --all-targets -- -D warnings` clean; full workspace test suite passes (rebased on top of #119).

Fixes #110

### DIFF
--- a/crates/sigstore-bundle/Cargo.toml
+++ b/crates/sigstore-bundle/Cargo.toml
@@ -15,7 +15,6 @@ native-tls = ["sigstore-tsa/native-tls", "sigstore-rekor/native-tls"]
 [dependencies]
 sigstore-types = { workspace = true }
 sigstore-crypto = { workspace = true }
-sigstore-merkle = { workspace = true }
 sigstore-tsa = { workspace = true, default-features = false }
 sigstore-rekor = { workspace = true, default-features = false }
 serde = { workspace = true }
@@ -26,3 +25,4 @@ hex = { workspace = true }
 
 [dev-dependencies]
 hex = { workspace = true }
+sigstore-merkle = { workspace = true }
--- a/crates/sigstore-bundle/src/error.rs
+++ b/crates/sigstore-bundle/src/error.rs
@@ -12,10 +12,6 @@ pub enum Error {
     /// JSON error
     #[error("JSON error: {0}")]
     Json(#[from] serde_json::Error),
-
-    /// Merkle proof error
-    #[error("Merkle proof error: {0}")]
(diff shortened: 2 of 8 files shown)


jku (approved):
This is great, makes it a lot easier to follow and moving the inclusion proof verification to verify seems just correct.

I was a bit worried seeing the "+328" but that was all comments :)

PR #125, opened 2026-06-16
ci: auto-open a PR to refresh embedded TUF data instead of filing an issue

### LINKED ISSUE #124: Embedded TUF data is not up-to-date
The Sigstore TUF repository contents have changed (or the update check failed): the TUF data embedded in the sigstore-trust-root crate can be updated.

Run
```
cargo run -p sigstore-trust-root --example update-embedded-roots
```
review the diff, and open a PR with the changes.

This issue was filed by [a scheduled workflow run](https://github.com/sigstore/sigstore-rust/actions/runs/27624175221).

wolfv:
Fixes #124's root cause for the future: the embedded-TUF-data check used to **file an issue** when the Sigstore TUF repositories changed, leaving someone to run the updater and open a PR by hand. This makes the workflow do both itself.

What changed in `.github/workflows/check-embedded-root.yml`:
- Run `cargo run -p sigstore-trust-root --example update-embedded-roots` in the job.
- Open (or update) a PR with the refreshed bytes via `peter-evans/create-pull-request` (pinned `# v8.1.1`), base `main`, branch `automation/update-embedded-tuf-data`, label `embedded-root-update`, `signoff: true`.
- A single instance that can't be refreshed no longer blocks the others — the updater step continues, and its outcome is reported in the PR body.

Note: creating PRs from Actions requires the repo setting **Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests"** to be enabled.

Heads-up surfaced while testing this: the GitHub Sigstore instance can't currently be refreshed by `main`'s `tough`-based updater — it rejects GitHub's root with an "Invalid key ID … calculated …" error (the key-ID recompute issue that #106 fixes). Production refreshes fine. Once #106 lands and the example runs on `sigstore-tuf`, the GitHub instance will refresh too.


jku:
> Note: creating PRs from Actions requires the repo setting Settings → Actions → General → "Allow GitHub Actions to create and approve pull requests" to be enabled.

looks like this can't be set with a PR to sigstore/community (the artisanal config format does not cover it). So we'd need an admin to enable that if we want it.

I think the PR approach is viable... reason I didn't originally go with that approach in sigstore-python was that I'm not 100% sure if we'd notice a somehow compromised workflow making a malicious PR. Maybe that's not a real risk: any usefully malicious changes would probably be obvious

PR #129, opened 2026-06-17
workflows: cleanup

jku:
* Use full version tags (no changes to the hashes)
* Switch one more toolchain setup to use rustup


wolfv (approved):
Thanks!

PR #130, opened 2026-06-17
Setup dependabot

jku:
I don't know what makes most sense for cargo/rust ecosystem but this is a basic setup that shouldn't be too annoying hopefully

@wolfv do you have opinions?


jku:
oops, I did not remember you had renovate already running -- where is that configured?

In any case this was too trigger happy, closing

wolfv:
I would also be happy with dependabot ... no strong feelings either way :)

PR #134, opened 2026-06-17
trust-root: fix update-embedded-roots when root is not rotated

wolfv:
The update-embedded-roots example read root.json back from the temp store after refresh, but the Updater only writes root.json when it rotates the root during the refresh. In steady state (embedded root already current) no rotation happens, the store has no root.json, and std::fs::read failed with ENOENT for every instance.

Fall back to the embedded root bytes when the store has no root.json: with no rotation, the freshest verified root is the one we bootstrapped from, so the embedded file stays byte-for-byte correct.

### DIFF
--- a/crates/sigstore-trust-root/examples/update-embedded-roots.rs
+++ b/crates/sigstore-trust-root/examples/update-embedded-roots.rs
@@ -112,8 +112,16 @@ async fn update_instance(instance: &Instance) -> Result<bool, Box<dyn std::error
     updater.refresh(now).await?;
 
     // Read the freshest verified root.json verbatim, to keep the embedded file
-    // byte-for-byte identical to what the repository serves.
-    let latest_root = std::fs::read(metadata_dir.path().join("root.json"))?;
+    // byte-for-byte identical to what the repository serves. The updater only
+    // writes `root.json` to the store when it rotates the root during this
+    // refresh; in the common steady state (embedded root already current) no
+    // rotation happens and the store has no `root.json`, so the freshest
+    // verified root is the embedded one we bootstrapped from.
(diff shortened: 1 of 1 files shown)


crates/sigstore-trust-root/examples/update-embedded-roots.rs
```diff
@@ -112,8 +112,16 @@ async fn update_instance(instance: &Instance) -> Result<bool, Box<dyn std::error
     updater.refresh(now).await?;
 
     // Read the freshest verified root.json verbatim, to keep the embedded file
-    // byte-for-byte identical to what the repository serves.
-    let latest_root = std::fs::read(metadata_dir.path().join("root.json"))?;
+    // byte-for-byte identical to what the repository serves. The updater only
+    // writes `root.json` to the store when it rotates the root during this
+    // refresh; in the common steady state (embedded root already current) no
+    // rotation happens and the store has no `root.json`, so the freshest
+    // verified root is the embedded one we bootstrapped from.
```
jku, line 119: so is this expected -- that we sometimes don't cache current metadata? 

I get that it's not necessary if the embedded root is always provided but feels like it would be logical to always store the current root in the cache (even if it it comes from embedded root)

I think I'll still approve this, I don't think it matters much.

jku (approved):

crates/sigstore-trust-root/examples/update-embedded-roots.rs
jku: I ended up fixing this in #136

jku (changes requested):
I *think* this won't be necessary after #136 -- I believe the cache is going to be complete even when the only root available is the embedded root

I suggest we try with that first

jku:
yup, I think we're ok now

PR #138, opened 2026-06-29
chore: release v0.10.0

wolfv:
## 🤖 New release

* `sigstore-types`: 0.9.0 -> 0.10.0
* `sigstore-crypto`: 0.9.0 -> 0.10.0
* `sigstore-merkle`: 0.9.0 -> 0.10.0
* `sigstore-tsa`: 0.9.0 -> 0.10.0
* `sigstore-tuf`: 0.9.0 -> 0.10.0 (✓ API compatible changes)
* `sigstore-trust-root`: 0.9.0 -> 0.10.0 (✓ API compatible changes)
* `sigstore-cache`: 0.9.0 -> 0.10.0
* `sigstore-rekor`: 0.9.0 -> 0.10.0
* `sigstore-bundle`: 0.9.0 -> 0.10.0
* `sigstore-oidc`: 0.9.0 -> 0.10.0
* `sigstore-fulcio`: 0.9.0 -> 0.10.0
* `sigstore-verify`: 0.9.0 -> 0.10.0 (✓ API compatible changes)
* `sigstore-sign`: 0.9.0 -> 0.10.0 (✓ API compatible changes)

<details><summary><i><b>Changelog</b></i></summary><p>

## `sigstore-types`

<blockquote>

### Other

- harden message-signature verification, drop digest leak ([#119](https://github.com/sigstore/sigstore-rust/pull/119))
- add `DigestBytes` for multi-algorithm digest encoding ([#109](https://github.com/sigstore/sigstore-rust/pull/109))
</blockquote>

## `sigstore-crypto`

<blockquote>

### Other

- harden message-signature verification, drop digest leak ([#119](https://github.com/sigstore/sigstore-rust/pull/119))
- support sha256/384/512 digests, fail closed on unsupported signing schemes, and add `KeyAlgorithm` ([#109](https://github.com/sigstore/sigstore-rust/pull/109))
- If a Fulcio OID is found, it must be parseable ([#108](https://github.com/sigstore/sigstore-rust/pull/108))
</blockquote>

## `sigstore-merkle`

<blockquote>

### Other

- fix overflow at maximum tree size ([#116](https://github.com/sigstore/sigstore-rust/pull/116))
</blockquote>

## `sigstore-tsa`

<blockquote>

### Other

- fail-closed if no root certs provided ([#114](https://github.com/sigstore/sigstore-rust/pull/114))
- Verify TSA response nonce ([#113](https://github.com/sigstore/sigstore-rust/pull/113))
[...]

PR #143, opened 2026-06-29
ci(release-plz): add DCO sign-off to release PR commit

wolfv:
## What

release-plz creates its release-PR commit via libgit2, so it has no way to add a `Signed-off-by` trailer itself (no `--signoff`, no `prepare-commit-msg` hook). This adds a CI step in the `release-plz-pr` job that, after the action pushes the branch, amends the commit to append a sign-off and force-pushes it.

## Details

- The sign-off identity is derived from the commit's own author (`%an`/`%ae`) so the trailer always matches the author and satisfies a DCO check regardless of which token created the commit.
- `--signoff` is idempotent — no duplicate trailer if one already exists.
- The amend/force-push targets the release branch (not `main`), so it does not re-trigger the `on: push: main` workflow.

This is the same trailer I added manually to #138; this makes it automatic going forward.

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -65,8 +65,30 @@ jobs:
         run: rustup toolchain install
 
       - name: Run release-plz
+        id: release-plz
         uses: release-plz/action@e8792575c7f2366cf6ff3ccc33ead9ace5b691c7 # v0.5.130
         with:
           command: release-pr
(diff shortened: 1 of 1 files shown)


jku (approved):
Always looks sketchy to include the github token in a URL... but I don't think there are alternatives and GitHub _should_ sanitize the logs.

PR #148, opened 2026-07-01
ci: auto-open a PR to refresh embedded TUF data

wolfv:
The scheduled embedded-data check filed an issue whenever the Sigstore TUF repositories changed, leaving a human to run the updater and open a PR by hand. Run the updater in the workflow itself and open (or update) a pull request with the refreshed bytes via peter-evans/create-pull-request, so the new roots arrive ready to review and merge.

A single instance that can't be refreshed no longer blocks the others: the updater step is allowed to continue, and its outcome is surfaced in the PR body so a partial refresh (e.g. an instance the current TUF client can't yet parse) is visible to reviewers.


jku (approved):
let's try this

PR #154, opened 2026-07-08
fix(verify): source SCT issuer from the verified cert chain

wolfv:
## Problem

The `sigstore-rust` job in the [root-signing-staging smoke tests](https://github.com/sigstore/root-signing-staging/actions/runs/28919686714) started failing on 2026-07-07 with:

```
Verification error: SCT signature verification failed: Verification error: ECDSA P-256 SHA-256 signature invalid
```

This coincides with the sigstore-public-good-**staging** move to multi-region. It is **not** a dependency change — it's a latent client bug exposed by staging infra.

## Root cause

SCT verification reconstructs the RFC 6962 `digitally-signed` structure. For precert entries that includes `issuer_key_hash = SHA-256(issuer SubjectPublicKeyInfo)`. When the bundle carried only the leaf certificate (no chain), the issuer was looked up in the trusted root by matching the subject **name**.

The multi-region rollout added a **second Fulcio intermediate** (in the staging trust root since 2026-06-12) that shares the subject `CN=sigstore-intermediate` with the pre-existing 2022 intermediate but has a **different key**:

| | Old (2022) | New region (2026-06-12) |
|---|---|---|
| Intermediate subject | `CN=sigstore-intermediate` | `CN=sigstore-intermediate` |
| Subject Key Identifier | `718630a6…` | `48612041…` |

`fulcio_certs()` returns them in trust-root order (old first), so name matching returned the **old** intermediate, the reconstructed `issuer_key_hash` was wrong, and the SCT signature check failed for every cert issued by the new region. The new region also runs a Static-CT/tile-based log (new log ID `1638fb66…`, present in the trust root), whose key `sigstore-rust` *does* find — hence "signature invalid" rather than "log ID not found".

Confirmed by reconstructing the RFC 6962 signed blob and verifying against the log's key: `INVALID` with the old intermediate's key hash, `VALID` with the new one (whose SKI matches the failing leaf's AKI).

## Fix

Take the SCT issuer from the **webpki-verified certificate path** built during chain verification, instead of re-resolving it by name. The verified path identifies the certificate that was cryptographically proven to have signed the leaf, so it is unambiguous by construction. This mirrors how sigstore-go and sigstore-python derive the SCT issuer from the verified chain parent.

`verify_certificate_chain` now returns the issuer's SPKI and the SCT step consumes it. The bespoke issuer-resolution logic (~120 lines) is removed — net **−100 lines**. SCT verification depends on the chain, which always runs when SCT does (`skip_certificate_chain()` clears both flags).

## Test

Adds a regression test using the **real staging trusted root and failing leaf bundle** (from the failing CI run). It asserts the fixture really contains two same-named intermediates, then drives the chain→issuer→SCT flow. Verified that under the old name-only logic the test fails with the exact production error, and passes with the fix. Full `sigstore-verify` suite (47 integration + 9 lib + doctests) and clippy are green.

## Note (unrelated)

The `cosign-old-version` canary failure (`ctfe public key not found`) in the same run is a separate, expected consequence of the same rollout: cosign 2.2 doesn't know the new Static-CT log's key. Not addressed here.


crates/sigstore-verify/src/verify_impl/helpers.rs
```diff
@@ -394,40 +402,129 @@ fn get_issuer_spki(
...
+        };
+        if &ca_cert.tbs_certificate.subject != issuer_name {
+            continue;
+        }
+        if let (Some(aki), Some(ski)) = (&authority_key_id, subject_key_id(&ca_cert)) {
+            if aki == &ski {
+                return spki_der(&ca_cert);
             }
         }
+        if name_match.is_none() {
```
jku: this might be a nit but in the case where aki != ski we should not name match:
```suggestion
        else if name_match.is_none() {
```

jku (changes requested):
This looks correct to me, thanks. Left one suggestion -- I don't think it's security relevant but reads slightly incorrect to me?

crates/sigstore-verify/src/verify_impl/helpers.rs
wolfv: sorry the PR was changed quite a bit to reuse the chain webpki/

jku (approved):
yeah this looks much better, thanks

PR #155, opened 2026-07-08
refactor(verify)!: make invalid verification policies unrepresentable

wolfv:
Follow-up to #154. Tightens the verification API so states that can't verify aren't expressible, removing runtime guards and vestigial fields.

### 1. `CertificatePolicy` enum replaces the `verify_certificate` + `verify_sct` bools
SCT verification consumes the issuer produced by chain verification (see #154). The two independent bools allowed the invalid combination "verify SCT but skip the chain", which was caught by a runtime error and an `Option` unwrap mid-verify.

```rust
pub enum CertificatePolicy {
    Skip,                          // verifies neither chain nor SCT
    Verify { verify_sct: bool },   // SCT flag only exists when the chain is verified
}
```

The invalid state is now unrepresentable; the issuer flows out of `verify_certificate_chain` as a plain value.

### 2. Remove `verify_timestamp` / `skip_timestamp()`
These were **never read**. A verified signing time (RFC3161 TSA timestamp or Rekor integrated time) is always required to validate short-lived Fulcio certs, so `skip_timestamp()` was a silent no-op advertising an unsafe option that doesn't — and shouldn't — exist.

This matches sigstore-python: its verifier establishes time from TSA/integrated-time sources, enforces `VERIFIED_TIME_THRESHOLD = 1` ("not enough sources of verified time"), and exposes **no** timestamp-skipping flag.

### 3. Remove `VerificationResult::success` (+ dead `failure()`)
Every failure returns `Err`, so `success` was always `true` on the `Ok` path — a trap for callers who checked the bool instead of the `Result`. `failure()` was never constructed.

## Breaking changes
- `VerificationPolicy`: `verify_certificate`/`verify_sct`/`verify_timestamp` fields and `skip_timestamp()` are gone; use the new `certificate: CertificatePolicy` field. `skip_certificate_chain()` and `skip_sct()` are unchanged.
- `VerificationResult`: no `success` field — a returned `Ok` already means success.

Updated call sites: examples, conformance runner, doctests, and tests.

## Test
Full `sigstore-verify` suite (9 lib + 47 integration + 4 doctests) passes; workspace builds `--all-targets`; fmt + clippy clean.

### DIFF
--- a/crates/sigstore-conformance/src/main.rs
+++ b/crates/sigstore-conformance/src/main.rs
@@ -265,17 +265,11 @@ fn verify_bundle(args: &[String]) -> Result<(), Box<dyn std::error::Error>> {
             let artifact_digest = Sha256Hash::try_from_slice(&digest_bytes)
                 .map_err(|e| format!("Invalid digest: {}", e))?;
 
-            let result = verify_with_key(artifact_digest, &bundle, &public_key, &trusted_root)?;
-            if !result.success {
-                return Err("Verification failed".into());
-            }
+            verify_with_key(artifact_digest, &bundle, &public_key, &trusted_root)?;
         } else {
             // It's a file path
             let artifact_data = fs::read(&artifact_or_digest)?;
-            let result = verify_with_key(&artifact_data, &bundle, &public_key, &trusted_root)?;
-            if !result.success {
-                return Err("Verification failed".into());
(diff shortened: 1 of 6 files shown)


jku (approved):
This all makes sense to me, thanks.

PR #157, opened 2026-07-08
fix(verify): tighten managed-key verification

wolfv:
## Summary
- reject certificate-backed bundles when caller-supplied public-key verification is requested
- treat `publicKey.hint` as the opaque, implementation-defined identifier specified by protobuf-specs; the signature establishes whether the supplied key is correct
- give managed-key verification its own policy, including an explicit unsafe opt-out from transparency-log inclusion verification
- expose managed-key verification through both the free function and `Verifier`
- report authenticated Rekor v1 integrated times and reject future timestamps
- replace the stale verification refactor with a focused change on current `main`

## Tests
- `cargo test -p sigstore-verify --all-targets`
- `cargo clippy --all-targets -p sigstore-verify -p sigstore-conformance -- -D warnings`

### DIFF
--- a/crates/sigstore-verify/src/verify.rs
+++ b/crates/sigstore-verify/src/verify.rs
@@ -3,14 +3,15 @@
 //! This module provides the main entry point for verifying Sigstore signatures.
 
 use crate::error::{Error, Result};
-use base64::Engine;
 use sigstore_bundle::validate_bundle_with_options;
 use sigstore_bundle::ValidationOptions;
 use sigstore_crypto::{parse_certificate_info, KeyAlgorithm, SigningScheme, VerificationKey};
 use sigstore_trust_root::TrustedRoot;
 
+use sigstore_types::bundle::VerificationMaterialContent;
 use sigstore_types::{
-    Artifact, Bundle, HashAlgorithm, Sha256Hash, Sha512Hash, SignatureContent, Statement,
+    Artifact, Bundle, HashAlgorithm, KindVersion, Sha256Hash, Sha512Hash, SignatureContent,
+    Statement,
 };
 
 /// How the signing certificate is verified.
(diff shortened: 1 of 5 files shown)


crates/sigstore-verify/src/verify.rs
```diff
@@ -235,191 +272,228 @@ impl Verifier {
...
+        (
+            VerificationMaterialContent::Certificate(_)
+            | VerificationMaterialContent::X509CertificateChain { .. },
+            VerificationMode::Certificate(policy),
+        ) => verify_certificate_bundle(artifact, bundle, policy, trusted_root),
+        (
+            VerificationMaterialContent::PublicKey { hint },
+            VerificationMode::PublicKey { public_key, policy },
+        ) => {
+            verify_public_key_hint(hint, public_key)?;
```
jku: I'm not sure the hint is actually strictly defined... Are you confident we should do this?

This code also exists in the TSA PR

crates/sigstore-verify/src/verify.rs
```diff
@@ -248,161 +276,293 @@ impl Verifier {
...
+/// - Skips certificate chain verification (no certificate present)
+/// - Skips identity/issuer verification
+fn verify_public_key_bundle(
+    artifact: Artifact<'_>,
+    bundle: &Bundle,
+    public_key: &DerPublicKey,
+    policy: &PublicKeyVerificationPolicy,
+    trusted_root: &TrustedRoot,
+) -> Result<VerificationResult> {
+    let result = VerificationResult::new();
```
jku: we don't populate `result` in this method (in practice I guess only integration time would be possible to populate)

crates/sigstore-verify/src/verify.rs
```diff
@@ -248,161 +276,293 @@ impl Verifier {
...
+    validate_bundle_shape(bundle, policy.verify_tlog)?;
+
+    // Verify transparency log entries (Merkle inclusion proofs, checkpoints,
+    // SETs) without certificate time validation
+    if policy.verify_tlog {
+        for entry in &bundle.verification_material.tlog_entries {
+            crate::verify_impl::tlog::verify_entry_inclusion(entry, trusted_root)?;
```
jku: This looks valid to me, I think the only additional check that could make sense might be the "ensure timestamp is not in future" handled by `validate_integrated_time` in the certificate case... but I'm not sure it's that important.

jku (dismissed):
Yeah I like this more. Left a couple of comments, but lgtm

jku (approved):

PR #159, opened 2026-07-27
refactor(trust-root)!: consolidate transparency key getters and align TimeRange semantics

wolfv:
Trust-root changes from the Trail of Bits week-1 review, plus follow-ups from PR review.

## TOB-SIGSTORE-5 — use the protobuf `TimeRange` semantics everywhere

`ValidityPeriod` and signing-config service validity represented the same protobuf-specs `TimeRange` message but implemented different behavior:

- trusted-root validity treated a missing `start` as unbounded;
- signing-config validity excluded the `end` timestamp;
- timestamps were repeatedly parsed at use time, and some verifier paths silently ignored parsing errors.

Both now use the shared public `TimeRange { start: Timestamp, end: Option<Timestamp> }` type. It implements the specified closed interval `[start, end]`, requires `start`, and treats a missing `end` as unbounded. Invalid or start-less ranges fail while parsing trust material rather than later during verification. `ServiceEndpoint` uses `TimeRange` directly rather than another alias.

All embedded trusted roots/signing configs and the checked conformance assets provide `start`.

## TOB-SIGSTORE-1 — return transparency-key `Keyring`s

The trust-root API exposed several overlapping Rekor/CTFE key getters with different collection shapes and identifier representations. `rekor_keys()` and `ctfe_keys()` now construct `sigstore_crypto::Keyring`s instead.

`Keyring` now owns the common lookup/filtering behavior:

- full SHA-256 key-ID lookup;
- four-byte checkpoint-hint lookup, including safe handling of hint collisions;
- closed validity-window lookup at a known production time;
- historical lookup that keeps expired keys when no exact production time is available.

Rekor checkpoint and SET verification, plus CT SCT verification, use these keyrings directly. SCT lookup checks the key's validity at the SCT's RFC 6962 timestamp; SET lookup checks validity at Rekor's integrated time when available.

The keyring indexes keys by the ID declared in the TUF-distributed trusted root. It no longer mixes checkpoint hints derived from the declared ID with SCT IDs recomputed from key bytes. Malformed non-SHA-256 IDs are rejected deterministically during keyring construction, independent of trusted-root array order.

CT roots may contain keys for algorithms other than the SCT being checked; incompatible candidates are filtered while constructing that SCT's keyring. Bare PKCS#1 RSA public keys used by Sigstore staging are also accepted in addition to SPKI keys.

## Managed-key hint interoperability

Public-key hints now accept all base64 forms permitted by protojson: standard or URL-safe alphabets, with or without padding. The non-protobuf `sha256:`/`SHA256:` prefix handling was removed.

## Breaking changes

- `rekor_keys()` now returns `Result<Keyring>`.
- `ctfe_keys()` now accepts the SCT `SigningScheme` and returns `Result<Keyring>`.
- `rekor_keys_with_hints()`, `ctfe_keys_with_ids()`, and the unused `has_rekor_key()` are removed.
- `ServiceEndpoint::valid_for` uses `TimeRange` directly; `ServiceValidityPeriod` is removed.
- `ValidityPeriod` is an alias for `TimeRange`, so `start` is a required `Timestamp` and `end` is an `Option<Timestamp>`.
- `fulcio_certs()`, TSA certificate getters, and TSA validity checks no longer return `Result`, because validity timestamps are parsed during deserialization.
- `TsaCertWithValidity` is now `(CertificateDer, Option<TimeRange>)` rather than a three-tuple of loose timestamps.

The semantic Rekor lookups `rekor_key_for_log()` and `rekor_key_for_log_at()` remain available.

## Validation

- `cargo test --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/sigstore-trust-root/src/signing_config.rs
+++ b/crates/sigstore-trust-root/src/signing_config.rs
@@ -70,18 +70,18 @@ pub struct ServiceValidityPeriod {
 }
 
 impl ServiceValidityPeriod {
+    /// Whether `time` falls within this validity period.
+    ///
+    /// Per the protobuf-specs `TimeRange` semantics the period is a closed
+    /// interval `[start, end]` (both bounds included); a missing `end` means
+    /// the period has started but has no known end.
+    pub fn contains(&self, time: Timestamp) -> bool {
+        crate::time_range::time_range_contains(self.start, self.end, time)
+    }
+
     /// Check if this period is currently valid
     pub fn is_valid(&self) -> bool {
-        let now = Timestamp::now();
-        if now < self.start {
-            return false;
-        }
-        if let Some(end) = self.end {
-            if now >= end {
-                return false;
-            }
-        }
-        true
+        self.contains(Timestamp::now())
     }
 }
 
@@ -334,5 +334,22 @@ mod tests {
             end: Some("2021-01-01T00:00:00Z".parse().unwrap()),
         };
         assert!(!expired_period.is_valid());
+    }
+
+    #[test]
+    fn test_service_validity_is_a_closed_interval() {
+        let start: Timestamp = "2020-01-01T00:00:00Z".parse().unwrap();
+        let end: Timestamp = "2021-01-01T00:00:00Z".parse().unwrap();
+        let period = ServiceValidityPeriod {
+            start,
+            end: Some(end),
+        };
+
+        // Both bounds are included per the TimeRange spec
+        assert!(period.contains(start));
+        assert!(period.contains(end));
+
+        assert!(!period.contains("2019-12-31T23:59:59Z".parse().unwrap()));
+        assert!(!period.contains("2021-01-01T00:00:01Z".parse().unwrap()));
     }
 }
--- a/crates/sigstore-trust-root/src/signing_config.rs
+++ b/crates/sigstore-trust-root/src/signing_config.rs
@@ -57,13 +57,6 @@ pub const SUPPORTED_FULCIO_VERSIONS: &[u32] = &[1];
 /// Expected media type for signing config v0.2
 pub const SIGNING_CONFIG_MEDIA_TYPE: &str = "application/vnd.dev.sigstore.signingconfig.v0.2+json";
 
-/// Validity period for a service.
-///
-/// A signing config's `validFor` is an instance of the protobuf-specs
-/// `TimeRange` message, so this is an alias for [`TimeRange`] — the same type
-/// the trusted root uses for its validity periods.
-pub type ServiceValidityPeriod = TimeRange;
-
 /// A service endpoint configuration
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -73,7 +66,7 @@ pub struct ServiceEndpoint {
     /// Major API version supported by this endpoint
     pub major_api_version: u32,
     /// Validity period for this endpoint
-    pub valid_for: ServiceValidityPeriod,
+    pub valid_for: TimeRange,
     /// Operator of this service
     #[serde(default)]
     pub operator: Option<String>,
@@ -302,13 +295,13 @@ mod tests {
 
     #[test]
     fn test_service_validity() {
-        let valid_period = ServiceValidityPeriod {
+        let valid_period = TimeRange {
             start: "2020-01-01T00:00:00Z".parse().unwrap(),
             end: None,
         };
         assert!(valid_period.is_valid());
 
-        let expired_period = ServiceValidityPeriod {
+        let expired_period = TimeRange {
             start: "2020-01-01T00:00:00Z".parse().unwrap(),
             end: Some("2021-01-01T00:00:00Z".parse().unwrap()),
         };
@@ -317,8 +310,8 @@ mod tests {
 
     #[test]
     fn test_service_endpoint_validity_is_a_time_range() {
-        // `ServiceValidityPeriod` is the protobuf-specs `TimeRange`; the
-        // containment semantics themselves are covered in `crate::time_range`.
+        // Service validity uses the protobuf-specs `TimeRange`; containment
+        // semantics are covered in `crate::time_range`.
         let endpoint: ServiceEndpoint = serde_json::from_str(
             r#"{
                 "url": "https://rekor.example.com",
--- a/crates/sigstore-trust-root/src/signing_config.rs
+++ b/crates/sigstore-trust-root/src/signing_config.rs
@@ -33,10 +33,9 @@
 //! let config = SigningConfig::from_json(SIGSTORE_PRODUCTION_SIGNING_CONFIG).unwrap();
 //! ```
 
-use jiff::Timestamp;
 use serde::{Deserialize, Serialize};
 
-use crate::{Error, Result};
+use crate::{time_range::TimeRange, Error, Result};
 
 /// Embedded production signing config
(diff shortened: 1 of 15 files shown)


crates/sigstore-trust-root/src/trusted_root.rs
tnytown: Why is this type distinct from `ServiceValidityPeriod`? They should both map to the same underlying `TimeRange` protobuf message.

crates/sigstore-verify/src/verify.rs
```diff
@@ -577,6 +601,39 @@ pub fn verify<'a>(
...
+        .strip_prefix("SHA256:")
+        .or_else(|| hint.strip_prefix("sha256:"))
+        .unwrap_or(hint);
+
+    let decoded = base64::engine::general_purpose::STANDARD
+        .decode(hint)
+        .or_else(|_| base64::engine::general_purpose::URL_SAFE_NO_PAD.decode(hint))
+        .map_err(|_| {
+            Error::Verification("public key hint is not a supported SHA-256 key hint".to_string())
+        })?;
```
tnytown: This decodes a base64 blob, optionally prefixed with `sha256:` or `SHA256:`. Two notes:

a) Is the prefix handling necessary? (The Protobuf specs leave the encoding of the key hint implementation-dependent.) If it is necessary, the test materials don't include a prefix.
b) The base64 decoding here does not account for padded encodings, [which is allowed in protojson](https://protobuf.dev/programming-guides/json/#:~:text=JSON%20value%20will%20be%20the%20data%20encoded%20as%20a%20string%20using%20standard%20base64%20encoding%20with%20paddings.%20Either%20standard%20or%20URL%2Dsafe%20base64%20encoding%20with/without%20paddings%20are%20accepted.).

crates/sigstore-verify/src/verify_impl/sct.rs
```diff
@@ -215,14 +215,19 @@ pub fn verify_sct(
...
-    })?;
+    let key = ct_keys
+        .iter()
+        .find(|key| key.computed_log_id().as_bytes() == log_id)
+        .ok_or_else(|| {
+            Error::Verification(format!(
+                "SCT log ID {:?} not found in trusted root CT logs",
+                hex::encode(log_id)
+            ))
+        })?;
```
tnytown: The `Keyring` type in `sigstore-crypto` encapsulates the lookup behavior you need here 🙂

crates/sigstore-trust-root/src/trusted_root.rs
```diff
@@ -262,53 +319,22 @@ impl TrustedRoot {
...
-    ///
-    /// Returns a vector of (key_hint, public_key) tuples where key_hint is
-    /// the first 4 bytes of the keyId from the log_id field.
+    /// Get all Rekor transparency log keys
     ///
     /// Keys whose `valid_for` window has not started yet are excluded.
     /// Expired keys are included because they are needed to verify log
     /// entries that were integrated while the key was valid.
-    pub fn rekor_keys_with_hints(&self) -> Result<Vec<(KeyHint, DerPublicKey)>> {
+    pub fn rekor_keys(&self) -> Result<Vec<LogKey>> {
```
tnytown: I may be missing something, but is there any reason why you're not using the `Keyring` type in `sigstore-crypto` here?

crates/sigstore-trust-root/src/trusted_root.rs
```diff
@@ -196,24 +197,80 @@ impl ValidityPeriod {
...
+    /// The log ID recomputed from the public key: the SHA-256 hash of the
+    /// DER-encoded key.
+    ///
+    /// SCT verification matches a certificate SCT's `LogID` against this
+    /// computed value rather than the declared [`Self::log_id`], so a trusted
+    /// root whose declared ID disagrees with its key material cannot redirect
+    /// the lookup.
+    pub fn computed_log_id(&self) -> Sha256Hash {
+        sigstore_crypto::sha256(self.public_key.as_bytes())
     }
```
tnytown: Keep in mind that a malicious trusted root can swap out any number of sensitive properties, including the underlying key bytes. We should be fine with trusting the log key ID as-is, although this does have marginal value as a defense-in-depth measure if you'd like to leave it in 🙂

tnytown (commented):
These are my initial impressions.

> A trusted root entry whose log ID is too short to yield a hint was therefore fatal only when it happened to precede the matching key in the tlogs array, and invisible when it followed it — the same trusted root could verify or fail purely on array order.

I'm not sure if we should put too much energy into handling malformed trustroot materials—I think there are many misencodings that can trigger unexpected behavior in sigstore-rust. These materials are trusted by virtue of their TUF distribution, and we should treat them as such.

crates/sigstore-trust-root/src/trusted_root.rs
jku: They should yes. There might be some minor differences in how missing periods or start times were expected to be handled (since trustroot did not originally contain them)... but I think combining them should be safe now

crates/sigstore-trust-root/src/trusted_root.rs
jku: yeah feels a bit odd that we're ok with the key hint being built from the log_id but then when verifying we compute the log id... if we don't trust the trusted_root data to be consistent with itself then we should probably validate that at deserialization time instead of using two potentially different values.

Seems fine though, we can go with this.

jku (commented):
I think this looks good, overall it seems to do exactly what the comment says and I can't find any real issues: 
* andrews comment about not second guessing trustedroot content holds but current code seems fine: I'm ok with this
* DCO is failing, can you fix that?

@tnytown do you want to have another look

crates/sigstore-trust-root/src/signing_config.rs
```diff
@@ -58,32 +57,12 @@ pub const SUPPORTED_FULCIO_VERSIONS: &[u32] = &[1];
...
-        }
-        true
-    }
-}
+/// Validity period for a service.
+///
+/// A signing config's `validFor` is an instance of the protobuf-specs
+/// `TimeRange` message, so this is an alias for [`TimeRange`] — the same type
+/// the trusted root uses for its validity periods.
+pub type ServiceValidityPeriod = TimeRange;
```
tnytown: Turbo nitpick: No need for a type alias here; using the `TimeRange` type directly should work :)

tnytown (commented):
My earlier comment about `Keyring` still applies; I would personally prefer to see filtering implemented there. Then, the key getters in `TrustedRoot` can construct and return `Keyring`; and downstreams can call getters on `Keyring` to get matching keys by hint and validity window. Current code should work though!

jku (approved):
This looks correct to me, thanks.

@wolfv can you comment on Andrews Keyring suggestion: should we file an issue to look at that later or do you disagree with it?

crates/sigstore-verify/src/verify_impl/sct.rs
wolfv: Doing this now.

crates/sigstore-verify/src/verify.rs
wolfv: Removed!

crates/sigstore-trust-root/src/trusted_root.rs
wolfv: Great point, doing this now! :)

crates/sigstore-trust-root/src/signing_config.rs
wolfv: Removed

PR #160, opened 2026-07-27
fix(verify): authenticate signature timestamps and bind log entries to the bundle

wolfv:
Verification-path changes from the Trail of Bits week-1 review, plus
follow-ups found while re-reviewing them.

### TOB-SIGSTORE-3 — authenticate `integratedTime` before trusting it

A v1 entry's `integratedTime` was used as the certificate validation time
without verifying the SET that covers it, so a tampered value was trusted
whenever tlog verification was skipped or reordered. The SET is now
verified before the timestamp is used; an entry that qualifies as a
timestamp source but whose SET does not verify is a hard error.

### TOB-SIGSTORE-4 — validate the certificate at *every* verified timestamp

`determine_validation_time` collapsed the bundle's timestamps into one
value (earliest TSA, else earliest integrated time). With multiple
timestamping authorities, one compromised TSA could contribute a backdated
timestamp that became the sole validation time, vouching for expired key
material. All verified timestamps are now returned and the certificate
chain and validity window are checked at each.

### TOB-SIGSTORE-2 — `skip_tlog()` → `skip_tlog_unsafe()`

Renamed, with the tradeoff documented on both the builder and the
`verify_tlog` field.

### Follow-up — a log entry was not bound to the bundle it travels in

`verify_tlog_consistency` — the check that each entry's body actually
describes this bundle's certificate, signature and artifact — was gated
behind `policy.verify_tlog`, while `determine_validation_times` consumed
tlog entries unconditionally.

A SET only proves that *some* entry was logged at the claimed time; it says
nothing about which bundle that entry belongs to. So under
`skip_tlog_unsafe()` a genuine, unmodified Rekor entry from an unrelated
signature could be spliced into a bundle and its SET-authenticated
`integratedTime` accepted as a verified signing time. The only constraint
was finding an entry integrated within the signing certificate's validity
window, which a public high-volume log makes easy. Reproduced against the
fixtures: a DSSE entry from `conda-attestation.sigstore.json` splices into
`cosign-v3-blob.sigstore.json`, passes SET verification, and fails only on
the unrelated certificate window.

The check now runs unconditionally. It takes no trusted root and performs
no log cryptography, so it is meaningful even for callers who cannot reach
a transparency log; the key-based verification path already called it
unconditionally.

### Follow-ups — smaller items

- Replace an `expect()` in the SCT branch with a verification error.
- Document that an unverifiable timestamp source is fatal even when the
  other source would have sufficed. This is stricter than sigstore-python,
  which discards failing sources and applies `VERIFIED_TIME_THRESHOLD = 1`,
  so bundles that verified before TOB-SIGSTORE-4 can fail after it.

### Behaviour changes worth flagging

- Bundles carrying log entries whose kind/version is unsupported now fail
  rather than being ignored along with the rest of tlog verification.
- An unverifiable SET is fatal despite a valid TSA timestamp, and vice
  versa.

### Testing

Both security properties are mutation-tested rather than merely asserted:
re-gating the consistency check kills the splice test, and
`validation_times.iter().take(1)` kills the every-timestamp test. Notably
the latter mutant — an exact revert of TOB-SIGSTORE-4 — passed the entire
suite before this PR adds that test.

Full workspace suite, `clippy -D warnings` and `cargo fmt --check` all pass
on this branch alone.

Touches `tests/verification_tests.rs`, so it will conflict textually (not
semantically) with the trust-root PR — whichever lands second needs a
trivial rebase of that one test file.


crates/sigstore-verify/src/verify.rs
```diff
@@ -415,9 +463,14 @@ impl Verifier {
...
-            crate::verify_impl::verify_tlog_consistency(bundle, &artifact)?;
-        }
+        //
+        //      This runs regardless of `policy.verify_tlog`. It needs no trusted
+        //      root and performs no log cryptography - it only checks that each
+        //      entry's body describes *this* bundle. Skipping it would let an
+        //      entry belonging to an unrelated signature ride along in the
+        //      bundle, and step (0) would accept that entry's SET-authenticated
+        //      integratedTime as a verified signing time.
+        crate::verify_impl::verify_tlog_consistency(bundle, &artifact)?;
```
jku, line 473: good catch

jku (approved):
Looks great!

PR #161, opened 2026-07-27
test(compat): enforce protobuf-specs wire compatibility in CI

wolfv:
sigstore-rust deliberately builds its public API on hand-written serde
types rather than the protobuf-generated ones, which leaves nothing
checking that those types stay compatible with the protobuf JSON wire
format. Flagged as a long-term recommendation in TOB-SIGSTORE-5.

Adds an unpublished `sigstore-protobuf-compat` crate whose tests validate
our `Bundle`, `TrustedRoot` and `SigningConfig` against the canonical
`sigstore_protobuf_specs` (prost + pbjson) types. For every embedded
trusted root and signing config, plus bundle fixtures spanning v0.1–v0.3,
DSSE and message-signature content, and certificate, chain and public-key
verification materials, each fixture is checked for four properties:

1. the fixture parses with the generated types,
2. our re-serialization is accepted by them — pbjson rejects unknown
   fields, so misnamed or invented fields fail here,
3. it decodes to a structurally identical protobuf value, using the
   generated types themselves as the equality oracle so proto3
   default-elision needs no fuzzy JSON comparison,
4. our types accept the canonical protojson the generated types emit.

Property 2's premise was verified directly rather than assumed: pbjson
does reject unknown fields, both at the top level and nested.

The existing `cargo test --workspace` CI job picks the crate up
automatically.

Also covers `VerificationMaterialContent::PublicKey` — the bare key
identifier used by managed and self-managed signing keys — which was the
one bundle shape none of the certificate-bearing fixtures exercised. It
reuses the existing `test_data/managed-key` fixture and adds no new asset.

---

Full workspace suite, `clippy -D warnings` and `cargo fmt --check` all pass
on this branch alone. Merges cleanly against the other two review PRs.


jku (approved):
test crate looks good: I'm approving but maybe we want to wait for the other PRs in case the verify changes are still changing 🤷

PR #162, opened 2026-07-27
ci: run conformance weekly

wolfv:
Run CI weekly so the `conformance-results` artifact is refreshed before its 7-day retention period expires.

This keeps sigstore-rust represented in the nightly conformance report even when `main` has no recent pushes.


jku (approved):
thanks

PR #164, opened 2026-07-30
fix(verify)!: remove clock skew, and use `jiff` Timestamp in the interfaces

wolfv:
Trail of Bits recommended us to remove the clock skew configuration:

> A trusted but compromised Rekor log returns a valid v1 SET whose integration time is 60
seconds in the future. The log neglects to incorporate the entry into the public view until the integration time has passed. A verifier using the default policy accepts a bundle including that SET during the 60 second grace period. Third party monitors do not see the signature event until the window elapses. This breaks the expectation that clients should only accept logged bundles.
> In this scenario, the future-dated integration time must fall within the signing certificate's validity window.

> ### Recommendations
> Short term, remove the configurable clock skew grace period.

I also prefer to use "concrete" types wherever possible, because a `i64` or `u64` doesn't let us know whether it is seconds, nanoseconds, unix epoch or whatever date format. Ideally, we should parse early, at the "boundary" from protocol to Rust code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1854,6 +1854,7 @@ version = "0.11.0"
 dependencies = [
  "base64",
  "hex",
+ "jiff",
  "serde",
  "serde_json",
  "sigstore-crypto",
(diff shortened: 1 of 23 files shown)


jku (approved):
Makes sense. 

I'm not that familiar with the various time crates but if you say jiff is something we can rely on in future then I don't see a problem exposing that.

PR #166, opened 2026-07-30
fix(verify): bind TSA timestamps to the authority that signed them (TOB-SIGSTORE-11)

wolfv:
**Merge coordination:** this overlaps PR #159 in the trust-root validity/accessor code. Merge #159 first, then rebase this branch and keep the per-authority TSA model from this PR.

Fixes **TOB-SIGSTORE-11 (Medium, Cryptography)**: "TSA timestamps can borrow an unrelated authority's validity interval."

RFC 3161 timestamp verification now binds each timestamp to the exact timestamp authority whose leaf certificate signed the CMS token, and consults **only that authority's** `valid_for` window for temporal authorization.

## Finding

`extract_tsa_timestamp` pooled roots, intermediates, and leaves from **all** configured TSAs into one `VerifyOpts` before calling `sigstore_tsa::verify_timestamp_response`, which returned only the authenticated time — not which signer authenticated it. The verifier then called `TrustedRoot::is_timestamp_within_tsa_validity`, an existential check: ANY configured authority with a covering (or absent) `valid_for` interval authorized the timestamp, regardless of which authority actually signed the token.

**Exploit scenario:** an organization keeps a retired TSA chain in its trust root for historical verification alongside a replacement TSA with a later epoch. A compromised retired signer issues a token outside its epoch; the token gets temporal authorization from the replacement TSA's interval. Trail of Bits reproduced this by adding an unrelated TSA entry with an *empty* cert chain and a covering interval, which rescued a token that was rejected under its actual signer's interval.

## Changes

Three commits: the fix, a breaking refactor that makes the vulnerable pattern unrepresentable, then a breaking API redesign (review follow-up) that makes the remaining verify-then-authorize split unrepresentable too.

**Commit 1 — the fix:**

- **`sigstore-trust-root`**: each timestamp authority's chain is now available split per authority (leaf, intermediates, root) plus its `valid_for` window, keeping the existing "not yet started" filtering. Authorities with an empty cert chain are skipped entirely — they can never authenticate a token and no longer contribute a validity window (this alone kills the ToB reproducer).
- **`sigstore-verify`**: `extract_tsa_timestamp` now tries each authority **in isolation** — only that authority's root as trust anchor, only its intermediates, only its leaf as the pinned signer certificate. On success, only that authority's `valid_for` is checked (absent window = unrestricted). A token that verifies under no configured authority, or only under authorities whose windows exclude its signed time, is a hard error. Every timestamp present in the bundle must still verify.
- **`sigstore-tsa`**: when `VerifyOpts::tsa_certificates` is non-empty, the CMS signer certificate must now match one of the pinned certificates. Previously, provided TSA certificates were merely *additional* signer candidates, so a token embedding its own certificate could be attributed to any authority sharing the same root CA — exactly the retired-signer/replacement-signer rotation case.

**Commit 2 — `refactor(trust-root)!`, moving the invariants to parse time / the type level:**

- `ValidityPeriod` now holds parsed `jiff::Timestamp`s: a trusted root with a malformed `validFor` timestamp is rejected by `from_json` instead of erroring deep inside verification. `contains`/`has_started_by` become infallible; the `Error::TimeParse` variant and every "returns an error if a `valid_for` timestamp is malformed" caveat disappear.
- The per-authority result is a `TsaAuthority` struct (named `leaf`/`intermediates`/`root`/`valid_for` fields) with an `authorizes(time)` method, so the "absent window = unrestricted; consult only the signing authority" rule lives in one place. `TrustedRoot::tsa_authorities()` is infallible.
- The APIs that enabled the vulnerability are **removed** rather than deprecated: `is_timestamp_within_tsa_validity`, `tsa_validity_for_time`, and the pooled `tsa_root_certs`/`tsa_intermediate_certs`/`tsa_leaf_certs`/`tsa_certs_with_validity` getters. TSA trust material can now only be consumed per-authority.
- `fulcio_certs`, `rekor_keys`, `ctfe_keys`, `ctfe_keys_with_ids` lose their now-impossible error paths and return values directly.

**Commit 3 — `refactor(tsa)!`, review follow-up: consume an authority as one unit**

Addresses the review note that the API still read like a footgun: callers had to remember to verify the CMS signature against an authority's chain and then *separately* check that same authority's `valid_for` window. A `with_validity_period`-style builder wouldn't fix it — the setter is forgettable, and certs and window would remain independently settable, so opts could still pair authority A's chain with authority B's window (the original misuse class).

The unit of verification is now the authority, not an options bag:

- **`sigstore-types`**: new shared `TsaAuthority` type (`leaf`/`intermediates`/`root`/`valid_for`), replacing the trust-root-local struct — no new dependencies, both crates already depend on `sigstore-types`.
- **`sigstore-tsa`**: the only public verification entry point is now `verify_timestamp_for_authority(token, sig, &TsaAuthority) -> Result<Timestamp>`. An `Ok` means the token was signed by that authority's leaf, chained to that authority's root, **and** falls inside that same authority's window — inseparable by construction, so a first-`Ok`-wins loop over authorities is correct automatically. An authenticated token outside its signer's window is the typed error `TimestampOutsideValidity { time }`, letting multi-authority callers distinguish "wrong authority" from "right authority, wrong time".
- The separable primitives are removed from the public API: `verify_timestamp_response`, `VerifyOpts`, and `TimestampResult` are now crate-private, and `TsaAuthority::authorizes` is gone.
- **`sigstore-verify`**: the authority-binding loop reduces to exactly the first-`Ok`-wins form, with unchanged error messages.

## Testing

New tests (all fixtures are real conformance / GitHub TSA artifacts):

- `sigstore-verify` (`verify_impl::helpers::tests::tsa_authority_binding`):
  - token accepted via its signing authority even with an unrelated authority present;
  - **finding repro**: signing authority's window excludes the token, unrelated authority's window covers it → rejected (both with a real second authority and with ToB's exact empty-cert-chain entry);
  - token signed by no configured authority → rejected; empty authority list → rejected;
  - authority without `valid_for` → accepted; inclusive window-boundary accept/reject;
  - multi-authority selection against GitHub's real trust root (five TSA entries with overlapping epochs).
- `sigstore-tsa`: pinned-signer accept (leaf) and reject (root pinned, embedded leaf signs); for the combined per-authority API: accept within window, the typed outside-window rejection carrying the authenticated time, and fail-closed when an authority's leaf did not sign the token (its window is never consulted).
- `sigstore-trust-root`: chain splitting per authority, empty-chain/not-yet-started skipping, `valid_for` carried through verbatim, and malformed `validFor` timestamps rejected at `from_json`.

`cargo fmt`, `cargo clippy -p sigstore-types -p sigstore-tsa -p sigstore-trust-root -p sigstore-verify --all-targets -- -D warnings`, and `cargo test --workspace` all pass; existing production/staging/GitHub TSA-path tests are unchanged and green.


jku (dismissed):
* Agree with removing old API, it does not seem salvageable
* PR looks good to me -- it's less (actual) code and seems easy to follow
* timestamp API still feels a bit footgun like -- you have to remember to first verify  signature, then separately check temporal authority

I'm going to approve since I couldn't come up with a really neat alternative design for the last point. This is what I came up with 

```rust
let opts = TsaVerifyOpts::new()
    .with_root(...)
    .with_intermediates(...)
    .with_tsa_certificates(...)
    .with_validity_period(authority.valid_for);
let result = verify_timestamp_response(ts_bytes, signature_bytes, opts)?;
// signature is verified, timestamp temporal validity is verified
```

which is cleaner but still requires caller to remember to use `with_validity_period` so I'm not sure it's that much better

crates/sigstore-tsa/src/verify.rs
```diff
@@ -316,6 +365,29 @@ fn verify_cms_signature(
...
+            .tsa_certificates
+            .iter()
+            .any(|cert| cert.as_ref() == signer_der.as_slice())
+        {
+            return Err(Error::SignatureVerificationError(
+                "CMS signer certificate does not match any provided TSA certificate".to_string(),
+            ));
+        }
+    }
+
```
tnytown: we can afford to be stricter here and make this simpler. match for the appropriate leaf certificate among those provided from the trust root (don't use certs from the CMS envelope); then perform verification with the cert from the trust root. no need for DER equality checks.

crates/sigstore-tsa/src/verify.rs
tnytown, line 389: can we use the new unified signature checking helpers here?

tnytown (commented):
looks like this encodes the correct behavior.

one thing i'll note is that the comments look LLM-generated. that's okay if you find them useful; but i find that comments are most useful to me when they're handwritten and reflect my understanding. otherwise, overly verbose comments can become a hindrance.

crates/sigstore-trust-root/src/trusted_root.rs
```diff
@@ -295,107 +293,39 @@ impl TrustedRoot {
...
-    /// Timestamp authorities whose `valid_for` window has not started yet are
-    /// excluded. Expired authorities are included because they are needed to
-    /// verify timestamps issued while they were valid.
-    pub fn tsa_leaf_certs(&self) -> Vec<CertificateDer<'static>> {
-        let now = Timestamp::now();
-        let mut leaves = Vec::new();
-        for tsa in &self.timestamp_authorities {
-            if !usable_for_verification(tsa.valid_for.as_ref(), now) {
+            let certs = &tsa.cert_chain.certificates;
+            let (Some(leaf), Some(root)) = (certs.first(), certs.last()) else {
```
tnytown: You might be able to use array destructuring here:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d434b4e2de650ebb0df345e55ee73401

crates/sigstore-tsa/src/verify.rs
wolfv: Updated in a391055. The signer is now selected exclusively from the authority leaf certificates supplied by the trusted root, and chain validation uses only that authority’s configured intermediates and root. CMS-embedded certificates are no longer used.

crates/sigstore-tsa/src/verify.rs
wolfv: Done in a391055. CMS signatures now go through `sigstore_crypto::verify_signature`; the TSA code only maps the CMS curve/digest identifiers to the shared `SigningScheme`.

crates/sigstore-trust-root/src/trusted_root.rs
wolfv: Done in a391055 using slice destructuring, with the single-certificate chain case handled explicitly.

jku (approved):
sure, let's go with this

PR #169, opened 2026-07-31
fix(sign): keep unbounded hashing and signing from starving the async executor (TOB-SIGSTORE-8)

wolfv:
## Summary

Fixes Trail of Bits finding **TOB-SIGSTORE-8** (severity: informational, category: denial of service): "Raw artifact signing performs potentially blocking actions in async context."

`Signer::sign` synchronously hashed and ECDSA-signed unbounded `Artifact::Bytes` input inside an async fn. On a large artifact this occupies an async executor worker thread for the whole duration, starving other tasks. `KeyPair::sign` also re-hashed the entire message inside aws-lc-rs before signing (and constructed a `SystemRandom`, which may perform I/O for entropy).

## Mechanism

Two complementary changes, chosen over `spawn_blocking` because `Signer::sign` borrows the artifact bytes (`Artifact<'a>`), and `spawn_blocking` requires `'static` — copying a potentially huge artifact to satisfy that would defeat the purpose:

1. **Prehashed ECDSA signing** (`sigstore-crypto`): new `KeyPair::sign_prehashed(Sha256Hasher) -> (Sha256Hash, SignatureBytes)`, built on `aws_lc_rs::EcdsaKeyPair::sign_digest` (available and algorithm-checked in the pinned aws-lc-rs 1.17). Signing the precomputed SHA-256 digest produces a signature that verifies equivalently to ECDSA-SHA256 over the raw message — the hashedrekord signature remains verifiable exactly as before — while making the signing step O(1) in the message size. It also drops the `SystemRandom` handle used by the full-message path.
2. **Cooperative hashing** (`sigstore-sign`): caller-controlled input (artifact bytes in `sign`, DSSE PAE bytes in `sign_raw_statement` and `create_dsse_rekor_entry`) is now hashed without materializing a second full-size PAE buffer in 64 KiB chunks with a yield to the executor between chunks. The yield is a ~20-line runtime-agnostic `yield_now` future (wake + `Pending` once), so no tokio runtime dependency is added to the library and the code works on any executor.

The public API is unchanged. `Signer::sign` and `Signer::sign_raw_statement` document the executor behavior. Statement JSON validation in `sign_raw_statement` stays synchronous: statements carry metadata (subject digests, predicates), not artifact contents, so they are realistically small.

## Testing

- New: `sha256_yielding` chunked hashing matches one-shot `sigstore_crypto::sha256` across boundary sizes (0, 1, chunk±1, multi-chunk).
- New: streaming DSSE PAE hashing matches the materialized PAE across chunk boundaries.
- New: prehashed signatures verify (via `VerificationKey::verify`) as ECDSA-SHA256 over the full raw message, in both `sigstore-crypto` and through the `sigstore-sign` helper.
- `cargo fmt`, `cargo clippy -p sigstore-sign -p sigstore-crypto --all-targets -- -D warnings`, `cargo test -p sigstore-sign -p sigstore-crypto` (all pass, incl. doc-tests), `cargo check --workspace`.


jku (commented):
I don't quite undertand why  the `sign_prehashed()` argument is Sha256Hasher and not the digest 
* we call finish_digest() anyway as the first thing in sign_prehashed
* when we call `sha256_yielding()` we always await anyway


It's possible I don't fully grasp the async rust model... I guess what I'm asking is :
_Does this design have an advantage over just adding `sigstore_crypto::sha256_yielding()` that has the same signature as `sigstore_crypto::sha256` but yields_

PR #170, opened 2026-07-31
fix(types)!: make multi-signature DSSE envelopes unrepresentable (TOB-SIGSTORE-9)

wolfv:
Require DSSE envelopes in Sigstore bundles to contain exactly one signature, ensuring payload, timestamp, and Rekor verification all use the same signature bytes.

`DsseEnvelope` now stores a single `DsseSignature`; serde preserves the one-element `signatures` wire format and rejects any other count while parsing. Verification no longer searches for any valid signature.

**BREAKING CHANGE:** Replace `DsseEnvelope.signatures: Vec<DsseSignature>` with `DsseEnvelope.signature: DsseSignature`. `DsseEnvelope::new` accepts one signature, and `Bundle::from_json` rejects DSSE envelopes without exactly one signature.

Addresses TOB-SIGSTORE-9.

### DIFF
--- a/crates/sigstore-sign/src/sign.rs
+++ b/crates/sigstore-sign/src/sign.rs
@@ -414,10 +414,10 @@ impl Signer {
         let dsse_envelope = DsseEnvelope::new(
             payload_type,
             payload,
-            vec![DsseSignature {
+            DsseSignature {
                 sig: signature.clone(),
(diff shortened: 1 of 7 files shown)


jku (approved):
I like this, LGTM

PR #173, opened 2026-08-04
fix(tuf): reverify cached delegated roles

wolfv:
## Summary

- reverify cached delegated metadata against the keys and threshold of the parent used by the current traversal
- retain role-name caching without allowing a previously verified delegating edge to authorize a different edge
- add an end-to-end regression test with two parents delegating the same role name under different keys

This addresses TOB-SIGSTORE-14.

## Testing

- `cargo test -p sigstore-tuf`
- confirmed the new regression test fails before the client-side revalidation and passes with the fix
- `cargo fmt --all -- --check`
- `git diff --check`


jku (approved):
This is a different solution than the one I used in python (I made TrustedMetadataSet keep track of the verified delegator-role pairs) but this looks fine to me as well

PR #174, opened 2026-08-04
fix(tuf): keep wildcards within path segments

wolfv:
Fix TUF delegation path matching so wildcards never cross `/` boundaries.

Delegation patterns and target paths are split into segments, then each pair is matched with `glob::Pattern`. This prevents `**` and other wildcard forms from recursively authorizing nested paths while retaining shell-style `*`, `?`, and character-class matching. Malformed patterns fail closed.

Includes regression coverage showing that `releases/**` matches `releases/package` but not `releases/stable/package`.

Addresses TOB-SIGSTORE-13.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2249,6 +2249,7 @@ dependencies = [
 name = "sigstore-tuf"
 version = "0.11.0"
 dependencies = [
+ "glob",
  "hex",
  "jiff",
  "reqwest",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,16 +145,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bstr"
-version = "1.12.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5cee35f73844aa3014bb606320a6c1f010249dbdf43342fe54b5a4f6a8ed4b79"
-dependencies = [
- "memchr",
- "serde_core",
-]
-
-[[package]]
 name = "bumpalo"
 version = "3.20.3"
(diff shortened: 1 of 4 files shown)


jku (commented):
I can't find issues in this implementation but implementing a tokenizer and all that seems like a lot, I wouldn't be surprised if there are issues.

How about this instead (I roughly translated the python implementation):
* split by directory separator to avoid any issues with `**`
* use plain glob on the parts


```rust
fn is_target_in_path_pattern(
    pattern: &str, target_path: &str,
) -> std::result::Result<bool, glob::PatternError> {
    let target_parts: Vec<&str> = target_path.split('/').collect();
    let pattern_parts: Vec<&str> = pattern.split('/').collect();

    // Verify all pattern segments parse cleanly to fail closed on malformed syntax.
    // Note: If support for multi-star segment wildcards like `release-**.json` or `***`
    // is ever required, consecutive stars within a segment could be collapsed to `*`
    // before compiling with `glob::Pattern`.
    let patterns: Vec<glob::Pattern> = pattern_parts
        .iter()
        .map(|seg| glob::Pattern::new(seg))
        .collect::<std::result::Result<Vec<_>, _>>()?;

    if target_parts.len() != patterns.len() {
        return Ok(false);
    }

    for (target_seg, pat) in target_parts.iter().zip(patterns.iter()) {
        if !pat.matches(target_seg) {
            return Ok(false);
        }
    }

    Ok(true)
}
```

I'm sure there might be pattern edge cases where it does not match (like the commented "***") but this handles the real world patterns and seems to fail closed

jku (approved):
Thanks

PR #175, opened 2026-08-04
fix(verify): handle canonical Rekor intoto entries

wolfv:
## Summary

- model Rekor v1 `intoto/0.0.2` canonical entries using `payloadHash`, `payloadType`, envelope hash, signatures, and signature public keys instead of requiring the proposed DSSE payload
- bind the canonical entry payload hash and payload type to the bundle DSSE envelope
- bind every logged signature and its certificate/public key to the bundle signature and signing certificate
- recognize `intoto/0.0.2` as a Rekor v1 integrated-time source rather than a Rekor v2 entry
- support SHA-512 in-toto subjects so the existing real `sigstore.js@2.0.0` provenance fixture exercises complete verification
- add positive and tampering tests for payload hash, payload type, signature, and signing certificate bindings

This addresses TOB-SIGSTORE-12.

## Regression coverage

Before the model change, the real canonical fixture failed while parsing its Rekor body with `missing field payload`. The complete verification flow now succeeds with that fixture.

## Testing

- `cargo test -p sigstore-rekor -p sigstore-verify`
- `cargo clippy -p sigstore-rekor -p sigstore-verify --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

### DIFF
--- a/crates/sigstore-verify/src/verify.rs
+++ b/crates/sigstore-verify/src/verify.rs
@@ -496,27 +496,24 @@ fn verify_dsse_artifact_binding(
         ));
     }
     let matches = match artifact {
-        Artifact::Bytes(bytes) => {
+        Artifact::Blob(bytes) => {
             let sha256 = hex::encode(sigstore_crypto::sha256(bytes));
             let sha512 = hex::encode(sigstore_crypto::sha512(bytes));
             statement.matches_sha256(&sha256) || statement.matches_sha512(&sha512)
         }
-        Artifact::Digest(hash) => match hash.len() {
-            32 => {
-                let digest = hex::encode(hash);
-                statement.matches_sha256(&digest)
+        Artifact::Digest(digest) => {
+            let value = hex::encode(digest.as_bytes());
+            match digest.algorithm() {
+                HashAlgorithm::Sha2256 => statement.matches_sha256(&value),
+                HashAlgorithm::Sha2512 => statement.matches_sha512(&value),
+                algorithm => {
+                    return Err(Error::Verification(format!(
+                        "unsupported pre-computed artifact digest algorithm: {}",
+                        algorithm
+                    )))
+                }
             }
-            64 => {
-                let digest = hex::encode(hash);
-                statement.matches_sha512(&digest)
-            }
-            length => {
-                return Err(Error::Verification(format!(
-                    "unsupported pre-computed artifact digest length: {}",
-                    length
-                )))
-            }
-        },
+        }
     };
 
     if !matches {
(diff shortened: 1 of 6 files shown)


crates/sigstore-verify/src/verify.rs
```diff
@@ -515,7 +507,37 @@ fn verify_dsse_artifact_binding(
...
+                let digest = hex::encode(hash);
+                statement.matches_sha256(&digest)
+            }
+            64 => {
+                let digest = hex::encode(hash);
+                statement
+                    .subject
+                    .iter()
+                    .any(|subject| subject.digest.sha512.as_deref() == Some(digest.as_str()))
+            }
```
jku: nit: Could add a `Statement.matches_sha512()`  so these didn't look like they do a different thing.

jku (dismissed):
I will rubber stamp this but I wonder if we could just drop the intoto support and focus on hashedrekord and dsse?

Anyway, lgtm

crates/sigstore-verify/src/verify.rs
wolfv: Done!

crates/sigstore-verify/src/verify_impl/rekor.rs
```diff
@@ -195,37 +210,215 @@ fn verify_intoto_v002(
...
+        }
+        VerificationMaterialContent::PublicKey { .. } => {
+            ExpectedVerifier::PublicKey(managed_key.ok_or_else(|| {
+                Error::Verification(
+                    "intoto Rekor signature cannot be bound without the managed public key"
+                        .to_string(),
+                )
+            })?)
         }
+    };
```
tnytown: key hint unused here? this is likely OK as it still needs to pass signature verification and the managed key is supplied to the verifier by the user

crates/sigstore-verify/src/verify_impl/rekor.rs
```diff
@@ -195,37 +210,215 @@ fn verify_intoto_v002(
...
+            match expected_verifier {
+                ExpectedVerifier::Certificate(bundle_cert) => {
+                    rekor_key.as_bytes() == certificate_public_key(bundle_cert)?.as_bytes()
+                }
+                ExpectedVerifier::PublicKey(bundle_key) => {
+                    rekor_key.as_bytes() == bundle_key.as_bytes()
+                }
+            }
+        }
+    };
```
tnytown: this control flow is confusing. there should be a way to get rid of the nested matches and `ExpectedVerifier` type here.

tnytown (commented):
looks like this behaves correctly, but i have a few code quality nits

crates/sigstore-verify/src/verify_impl/rekor.rs
wolfv: Right. The hint is only locator metadata; the caller-supplied managed key is the trust input, and the DSSE signature still has to verify with it. I left the hint out of this consistency check intentionally.

crates/sigstore-verify/src/verify_impl/rekor.rs
wolfv: Simplified in the latest branch: both bundle and Rekor verifier forms are normalized to a DER public key, then compared once. This removes `ExpectedVerifier` and the nested matches.

jku (approved):

PR #176, opened 2026-08-04
feat: complete and harden Rekor v2 support

wolfv:
## Summary

Complete and harden Rekor v2 support across the client, signer, bundle validator, and verifier.

This deliberately removes the obsolete `dsse/0.0.2` API and the lossy v2-to-v1 response conversion. Rekor v2 supports DSSE envelopes through `hashedrekord/0.0.2`, and its `TransparencyLogEntry` response is now stored directly in the bundle.

### Rekor client

- make API version an explicit client property (`new_v2`, `public_v2`, `staging_v2`, and builder selection)
- reject version-specific operations locally when used with the wrong client
- apply a 30-second HTTP timeout, satisfying Rekor v2's >=20-second write requirement
- return protobuf-compatible `TransparencyLogEntry` values directly from v2 writes
- strictly reject malformed, non-hashedrekord, nonzero-integrated-time, SET-bearing, proof-less, or checkpoint-less v2 responses
- add v2 checkpoint, hash-tile, and entry-bundle read APIs
- parse checkpoints and return coordinate-preserving tile/entry-bundle wrappers
- support certificate and DER public-key verifiers with typed P-256 key details

### Removed code

- remove `DsseEntryV2`, `DsseRequestV002`, and `create_dsse_entry_v2`
- remove the duplicate `LogEntryV2`/`InclusionProofV2` models
- remove duplicated v2 response conversion and every `unwrap_or_default()` numeric fallback

### Signing

- submit DSSE PAE digests as `hashedrekord/0.0.2`
- carry v2 responses directly into bundles instead of round-tripping through v1 hex encodings
- require an RFC 3161 TSA for Rekor v2 configurations
- derive the submitted key-details value from the selected signing scheme
- correct documentation claiming the basic default was v2 (it is v1)

### Bundle validation and verification

- recognize only `hashedrekord/0.0.2` as Rekor v2
- follow the Rekor v2 specification by ignoring unauthenticated duplicate `inclusionProof.logIndex`, `treeSize`, and `rootHash` fields
- use the top-level log index and signed checkpoint tree size/root for v2 Merkle verification
- retain v1's duplicate-field consistency checks
- handle checkpoints containing multiple log/witness signatures without letting an invalid earlier matching signature suppress a later valid log signature

### Documentation

- replace the stale Rekor README with a v1/v2 capability matrix and write/read examples
- document the absence of v2 search and online-verification APIs
- document that parsed checkpoints still require signature verification

## Test-first validation

The new regressions failed before implementation:

- `cargo test -p sigstore-rekor --test v2_client` failed to compile because the versioned/read APIs and typed request constructors did not exist.
- `rekor_v2_ignores_untrusted_duplicate_inclusion_proof_fields` failed because structural validation trusted the duplicate proof root.
- `rekor_v2_accepts_a_valid_log_signature_after_an_invalid_matching_signature` failed on the first invalid matching signature instead of checking the later valid signature.

## Testing

```console
cargo test -p sigstore-rekor -p sigstore-bundle -p sigstore-sign -p sigstore-verify
cargo clippy -p sigstore-rekor -p sigstore-bundle -p sigstore-sign -p sigstore-verify --all-targets --all-features -- -D warnings
cargo test --workspace --all-features --no-run
cargo fmt --all -- --check
git diff --check
```

All pass locally. Existing live Rust/Cosign v2 interoperability coverage remains in `.github/workflows/interop.yml`.

## Breaking changes

This intentionally removes the never-standardized `dsse/0.0.2` submission API and changes `create_entry_v2` to return `TransparencyLogEntry` rather than the v1 `LogEntry` compatibility type.

## Cross-implementation references

The implementation and compatibility tests are cross-checked against pinned versions of the other Sigstore clients:

- [`sigstore-python` Rekor v2 request construction](https://github.com/sigstore/sigstore-python/blob/71b63df1aea910fbfe5ba0fb1dc1199a27f56b25/sigstore/_internal/rekor/client_v2.py) for direct `TransparencyLogEntry` responses and DSSE-as-hashedrekord submission.
- [`sigstore-python` Rekor v2 bundle binding](https://github.com/sigstore/sigstore-python/blob/71b63df1aea910fbfe5ba0fb1dc1199a27f56b25/sigstore/verify/verifier.py#L429-L445) for PAE digest, signature, certificate, and key-details checks.
- [`sigstore-go` Rekor v2 submission](https://github.com/sigstore/sigstore-go/blob/22d3691c7b8e0c5530fae3c05577690bfef5cd00/pkg/sign/transparency.go#L154-L235) for algorithm-derived prehashing and direct protobuf response storage.
- [`sigstore-go` Rekor v2 verification](https://github.com/sigstore/sigstore-go/blob/22d3691c7b8e0c5530fae3c05577690bfef5cd00/pkg/verify/tlog.go#L112-L151) for tile-log checkpoint/inclusion verification distinct from v1.

Pinned upstream sigstore-python message-signature and DSSE Rekor v2 fixtures are now included with provenance under `crates/sigstore-verify/test_data/upstream/sigstore-python/`; both are verified by the Rust test suite.

## Protobuf-spec compatibility

The non-published `sigstore-protobuf-compat` crate now also checks Rekor v2 directly against `sigstore_protobuf_specs` 0.5.1:

- certificate-backed `HashedRekordRequestV002` / `CreateEntryRequest`
- public-key-backed `HashedRekordRequestV002` / `CreateEntryRequest`
- both imported sigstore-python Rekor v2 bundles as canonical protobuf `Bundle` JSON

Each check verifies fixture acceptance by generated pbjson types, acceptance of the Rust reserialization, structural equality after a Rust round trip, and Rust acceptance of canonical JSON emitted by the generated types. This catches misspelled/invented fields, wrong oneof shapes, incorrect bytes encoding, default-elision incompatibility, and dropped protobuf data.


crates/sigstore-bundle/src/validation.rs
```diff
@@ -220,30 +220,39 @@ fn validate_inclusion_proof_structure(bundle: &Bundle) -> Result<()> {
...
-            if checkpoint.root_hash != proof.root_hash {
-                return Err(Error::Validation(
-                    "inclusion proof root hash does not match checkpoint root hash".to_string(),
-                ));
+            let is_v2 =
+                entry.kind_version.kind == "hashedrekord" && entry.kind_version.version == "0.0.2";
+            if is_v2 {
+                let leaf_index = entry
+                    .log_index
+                    .as_u64()
```
tnytown: i wonder if it's possible to deserialize and store the expected representation instead of having getters on types containing the original format. this pattern appears a lot in the codebase where we're effectively performing deserialization outside of the deserialization layer

crates/sigstore-verify/src/verify_impl/tlog.rs
tnytown, line 1: i think `integratedTime` from a V2 bundle ends up being passed into `VerificationResult`. not a critical mistake because it's not used in verification, but might end up confusing downstream users. maybe we can remove the `VerificationResult.integrated_time` field?

crates/sigstore-rekor/src/client.rs
```diff
@@ -1,17 +1,69 @@
...
+
+impl RekorV2Tile {
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.bytes
+    }
+
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.bytes
+    }
+}
```
tnytown: are these getters necessary? the underlying `bytes` field is public

crates/sigstore-rekor/src/client.rs
```diff
@@ -1,17 +1,69 @@
...
+
+impl RekorV2EntryBundle {
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.bytes
+    }
+
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.bytes
+    }
+}
```
tnytown: same for these getters

crates/sigstore-verify/src/verify_impl/hashedrekord.rs
```diff
@@ -121,9 +125,46 @@ fn validate_certificate_match(
...
+            Error::Verification(
+                "Rekor verifier cannot be bound without the managed public key".to_string(),
+            )
+        })?;
+        let rekor_key = match body {
+            RekorEntryBody::HashedRekordV001(rekord) => rekord
+                .spec
+                .signature
+                .public_key
+                .to_public_key()
```
tnytown: the name of this conversion function is confusing

tnytown (commented):
just a few code quality nits that caught my eye; looks like this behaves correctly

crates/sigstore-bundle/src/validation.rs
wolfv: Agreed on deserializing into semantic types as early as possible. Commit 52026a1 now makes log indices and proof tree sizes validated non-negative types at deserialization, removing the downstream conversions. A fully protocol-specific normalized v1/v2 entry representation would require a broader public bundle-model change, so I have left that part for a focused follow-up.

crates/sigstore-verify/src/verify_impl/tlog.rs
wolfv: Fixed in c9ac745. Rekor v2 `integratedTime` is now ignored because it is not authenticated by a v2 SET; it is neither validated nor exposed through `VerificationResult`. I kept the field for authenticated Rekor v1 integrated times and added a regression test for a v2 entry carrying a bogus value.

crates/sigstore-rekor/src/client.rs
wolfv: Removed both convenience getters in c9ac745; callers now use the public `bytes` field directly.

crates/sigstore-rekor/src/client.rs
wolfv: Removed these as well in c9ac745.

crates/sigstore-verify/src/verify_impl/hashedrekord.rs
wolfv: Renamed the fallible PEM conversions to `parse_public_key` / `parse_certificate` in c9ac745 to make the parsing operation explicit.

crates/sigstore-rekor/src/client.rs
```diff
@@ -445,6 +353,148 @@ impl RekorClient {
...
+        Self {
+            url: url.into().trim_end_matches('/').to_string(),
+            client: build_http_client(timeout),
+        }
+    }
+
+    /// Create a client for the public Sigstore Rekor v2 instance.
+    pub fn public() -> Self {
+        Self::new(RekorApiVersion::V2.default_url())
+    }
```
jku, line 388: I wish we didn't have these methods especially for v2 -- leaving URLs visible in the examples would make it clearer that the caller is responsible for the URL and that `V2.default_url()` is unlikely to point to a usable log 100% of the time.

jku (commented):
I think this looks really good. I can't find real problems and it contains multiple major improvements. Left one comment on the "default URLs", but I don't think that's a blocker as it matches the rest of the low level API.

Reviewing PRs this large is kind of painful even with the separate commits. If the pr-stacks worked for you maybe that would work for this sort of thing in future

jku (approved):

PR #179, opened 2026-08-14
fix(trust-root)!: verify cached TUF repositories offline

wolfv:
## Summary

Offline TUF is now exactly one repository-access policy: **make no network requests and run the normal TUF verification workflow against the local store**.

- Replaces raw cache reads with `StoreRepository`.
- Re-verifies root rotation, timestamp, snapshot, targets, expiry, versions, and target hashes.
- Removes `OfflineMode` and automatic embedded fallback; embedded trust material remains a separate explicit source.
- Adds `TrustedRoot::from_tuf_at`, `SigningConfig::from_tuf_at`, and `fetch_trust_material_at` so applications can choose the metadata validation time.
- Existing entry points continue to validate at the current time.

A historical validation time supports the deliberate “offline, possibly expired now, but valid at the last refresh” policy without adding a second offline mode. It must come from application-owned state, not cache metadata or filesystem timestamps.

## Security

The writable cache is always treated as untrusted input. A missing, incomplete, expired-at-the-selected-time, or tampered cache is an error. Raw target bytes are never returned.

## Tests

Covers a fully verified cache, target tampering, missing metadata, absent/disabled cache, current-time expiry, application-selected historical validation, and the TOB-SIGSTORE-10 reproduction.

## Stack

1. **This PR:** verified offline repository and caller-selected validation time
2. #180: independent bootstrap trust policy, including explicit unsafe cached-root trust

Supersedes #165.

BREAKING CHANGE: `TufConfig::offline()` no longer takes `OfflineMode`, and automatic embedded fallback has been removed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2208,6 +2208,7 @@ dependencies = [
  "rustls-pki-types",
  "serde",
  "serde_json",
+ "sha2",
  "sigstore-crypto",
  "sigstore-tuf",
  "sigstore-types",
--- a/crates/sigstore-trust-root/Cargo.toml
+++ b/crates/sigstore-trust-root/Cargo.toml
@@ -57,5 +57,7 @@ required-features = ["tuf"]
 
 [dev-dependencies]
 tempfile = { workspace = true }
+# For building signed TUF fixtures in offline-cache tests
+sha2 = { workspace = true }
 # For testing
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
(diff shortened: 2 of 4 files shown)


jku (approved):
This seems correct, and I agree with the idea: exposing "now" lets the application decide when TUF needed to be up-to-date. Obviously now the application has the problem of storing the date safely... but there's definitely use case where this is a much better solution.

I'm not sure if it's good to keep adding API to sigstore-trust-root (instead of just breaking API)... But the old methods look pretty simple wrappers so are likely fine.

PR #180, opened 2026-08-14
feat(trust-root)!: make TUF bootstrap trust explicit

wolfv:
## Summary

Separates the **source of bootstrap trust** from whether the repository is online or offline.

Introduces:

```rust
pub enum TufBootstrap {
    TrustedRoot(Vec<u8>),
    UnsafeCachedRoot,
}
```

- Production, staging, and GitHub configurations use their embedded trusted roots.
- Custom repositories must explicitly select a bootstrap policy at construction.
- `TrustedRoot` is the safe/default choice.
- `UnsafeCachedRoot` preserves the generic-client workflow where a user establishes trust once and later commands use the cached root.
- The selected policy applies identically online and offline.
- `TufConfig` fields are private so callers cannot construct contradictory or implicit trust configurations.
- Generic sign/verify examples now make their cached-root trust decision visible.

## Security model

`UnsafeCachedRoot` is intentionally loud: anyone who can replace the writable cache can replace the root of trust and control the repository. Metadata and targets are still fully TUF-verified against that root, but the root's origin is not protected. This is an application-selected trust policy, not an offline behavior.

## Stack

1. #179: verified offline repository and caller-selected validation time
2. **This PR:** independent bootstrap trust policy

Supersedes #171 and directly addresses the generic-client concern raised on #165.

BREAKING CHANGE: `TufConfig::custom` now requires a `TufBootstrap` argument, `with_root` is removed, and `TufConfig` fields are private.

### DIFF
--- a/crates/sigstore-sign/examples/sign_blob.rs
+++ b/crates/sigstore-sign/examples/sign_blob.rs
@@ -129,7 +129,10 @@ async fn main() {
     // Create signing context with appropriate API version
     let tuf_config = if let Some(ref url) = instance {
         println!("  Using: custom instance ({})", url);
-        let config = sigstore_trust_root::tuf::TufConfig::custom(url);
+        let config = sigstore_trust_root::tuf::TufConfig::custom(
+            url,
+            sigstore_trust_root::tuf::TufBootstrap::UnsafeCachedRoot,
+        );
         sigstore_trust_root::SigningConfig::from_tuf(config)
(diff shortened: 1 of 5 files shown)


jku (approved):
The compromise feels correct to me: It's still possible to establish trust to a random tuf repo dynamically and use it without constantly jumping through hoops... But the application must make a conscious effort to do so.

PR #183, opened 2026-08-14
fix(sign): keep unbounded hashing and signing from starving the async executor

wolfv:
## Summary

Fixes TOB-SIGSTORE-8: unbounded artifact hashing and ECDSA signing no longer monopolize an async executor thread or hash the same input twice.

- Hashes blobs and DSSE PAE input incrementally in 64 KiB chunks.
- Yields between chunks without coupling the library to a runtime.
- Adds prehashed ECDSA signing, preserving signatures equivalent to ECDSA-SHA256 over the complete message.
- Avoids materializing a second full-size DSSE PAE buffer.

## Stack

1. **This PR:** incremental/prehashed signing foundation
2. #184: typed blob/digest artifact API
3. #185: synchronous and asynchronous reader APIs

Supersedes #169.


jku (approved):
sure, looks reasonable

PR #184, opened 2026-08-14
refactor(artifact)!: make artifact digests typed

wolfv:
## Summary

Makes artifact inputs explicit and type-safe before 1.0:

```rust
pub enum Artifact<'a> {
    Blob(&'a [u8]),
    Digest(ArtifactDigest),
}
```

`ArtifactDigest` carries its `HashAlgorithm` and validates the digest length at construction. Verification therefore rejects algorithm mismatches instead of interpreting arbitrary bytes according to whichever bundle field is inspected later.

Signing now accepts SHA-256 digest input as documented, signs it directly in prehashed mode, and uses the same value in the hashedrekord entry. Supplying a digest explicitly means the caller—not the library—is asserting the artifact identity.

## Stack

1. #183: incremental/prehashed signing foundation
2. **This PR:** typed blob and digest inputs
3. #185: synchronous and asynchronous reader APIs

BREAKING CHANGE: digest inputs must use `ArtifactDigest`; untyped `DigestBytes` conversions are removed and `Artifact::Bytes` is renamed `Artifact::Blob`.

### DIFF
--- a/crates/sigstore-crypto/src/signing.rs
+++ b/crates/sigstore-crypto/src/signing.rs
@@ -3,12 +3,13 @@
 use crate::error::{Error, Result};
 use crate::hash::Sha256Hasher;
 use aws_lc_rs::{
+    digest::{Digest, SHA256},
     rand::SystemRandom,
     signature::{EcdsaKeyPair, KeyPair as AwsKeyPair, ECDSA_P256_SHA256_ASN1_SIGNING},
 };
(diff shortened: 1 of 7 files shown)


jku (approved):
lgtm -- although I actually reviewed the version 20 mins ago (before the rebase): I'm assuming it was a clean rebase, let me know if there were actual changes

PR #185, opened 2026-08-14
feat(artifact)!: stream artifacts from sync and async readers

wolfv:
## Summary

Adds constant-memory artifact streaming to signing and verification.

### Signing

- `Signer::sign(...)` keeps accepting raw bytes or a typed SHA-256 digest
- `Signer::sign_reader(...)` streams a blocking `std::io::Read` to EOF
- `Signer::sign_async_reader(...)` streams a runtime-independent `futures_io::AsyncRead`

The artifact is hashed before the Fulcio certificate is requested, so large inputs no longer eat into the certificate's lifetime. The signer configuration is validated before any hashing starts.

### Verification

- `Verifier::verify(...)` and the free `verify(...)` keep accepting raw bytes or a typed digest
- `Verifier::verify_reader(...)` and `Verifier::verify_async_reader(...)` stream the artifact
- Managed-key verification mirrors this with `verify_with_key`, `verify_with_key_reader` and `verify_with_key_async_reader`
- The free `verify_with_key(...)` is now a thin wrapper over the `Verifier` method instead of a second copy of the verification flow

Readers are consumed exactly once. A new crate-private `PreparedArtifact` hashes the input in a single pass with every algorithm the bundle can ask for, and message-signature checks, DSSE subject binding, and Rekor consistency checks all draw on those digests without retaining or rereading the artifact. DSSE bundles hash only with the algorithms their in-toto subjects use; message-signature bundles hash SHA-256 for hashedrekord, the declared `messageDigest` algorithm, and, for streamed input, the digest the signing scheme consumes (derived from the bundle's certificate or the supplied managed key up front).

Blob input remains available for schemes such as Ed25519 that need the original message. Reader and digest input fail closed when the selected scheme does not support prehashed verification. Synchronous reader methods document that reads block the calling thread.

### Shared hashing helpers

`sigstore_crypto` gains `hash_reader`, `hash_reader_yielding` and `hash_async_reader`, which feed one pass over a reader into any set of hashers. The async variants yield to the executor after every 64 KiB chunk (TOB-SIGSTORE-8), and the signer's cooperative in-memory hashing now goes through the same code. The library crates depend on `futures-io` only; `futures` remains a dev-dependency.

CLI examples stream files. Tests verify real hashedrekord and DSSE bundles through both sync and async readers, and unit tests pin the per-bundle digest selection.

## Stack

1. #183: incremental/prehashed signing foundation
2. #184: typed blob and digest inputs
3. **This PR:** synchronous/asynchronous reader sources

### DIFF
--- a/crates/sigstore-crypto/src/hash.rs
+++ b/crates/sigstore-crypto/src/hash.rs
@@ -51,7 +51,7 @@ impl ArtifactHasher {
 
     pub fn finalize(self) -> ArtifactDigest {
         ArtifactDigest::new(self.algorithm, self.context.finish().as_ref())
-            .expect("cryptographic backend returned the algorithm's documented digest length")
+            .expect("backend digest length does not match the algorithm's documented length")
     }
 }
 
--- a/crates/sigstore-crypto/src/hash.rs
+++ b/crates/sigstore-crypto/src/hash.rs
@@ -149,7 +149,7 @@ pub fn sha256_reader(reader: impl Read) -> io::Result<Sha256Hash> {
 pub fn hash_reader<H: HashUpdate>(mut reader: impl Read, hashers: &mut [H]) -> io::Result<()> {
     let mut buffer = [0_u8; HASH_CHUNK_SIZE];
     loop {
-        let read = reader.read(&mut buffer)?;
+        let read = read_retry(&mut reader, &mut buffer)?;
         if read == 0 {
             return Ok(());
         }
@@ -202,6 +202,15 @@ pub async fn hash_async_reader<H: HashUpdate>(
     }
 }
 
+fn read_retry(reader: &mut impl Read, buf: &mut [u8]) -> io::Result<usize> {
+    loop {
+        match reader.read(buf) {
+            Err(e) if e.kind() == io::ErrorKind::Interrupted => continue,
+            result => return result,
+        }
+    }
+}
+
 /// Adapter presenting a blocking [`Read`] as an always-ready [`AsyncRead`].
 struct BlockingReader<R>(R);
 
@@ -215,7 +224,7 @@ impl<R: Read> AsyncRead for BlockingReader<R> {
         _cx: &mut TaskContext<'_>,
         buf: &mut [u8],
     ) -> Poll<io::Result<usize>> {
-        Poll::Ready(self.0.read(buf))
+        Poll::Ready(read_retry(&mut self.0, buf))
     }
 }
 
@@ -224,25 +233,16 @@ impl<R: Read> AsyncRead for BlockingReader<R> {
 /// Runtime-agnostic equivalent of `tokio::task::yield_now()`: the future
 /// wakes its own waker and returns `Pending` on the first poll, then `Ready`.
 fn yield_now() -> impl Future<Output = ()> {
-    struct YieldNow {
-        yielded: bool,
-    }
-
-    impl Future for YieldNow {
-        type Output = ();
-
-        fn poll(mut self: Pin<&mut Self>, cx: &mut TaskContext<'_>) -> Poll<()> {
-            if self.yielded {
-                Poll::Ready(())
-            } else {
-                self.yielded = true;
-                cx.waker().wake_by_ref();
-                Poll::Pending
-            }
+    let mut yielded = false;
+    std::future::poll_fn(move |cx| {
+        if yielded {
+            Poll::Ready(())
+        } else {
+            yielded = true;
+            cx.waker().wake_by_ref();
+            Poll::Pending
         }
-    }
-
-    YieldNow { yielded: false }
+    })
 }
 
 #[cfg(test)]
@@ -326,14 +326,11 @@ mod tests {
     }
 
     fn futures_task_noop_waker() -> std::task::Waker {
-        use std::task::{RawWaker, RawWakerVTable, Waker};
-        fn noop(_: *const ()) {}
-        fn clone(p: *const ()) -> RawWaker {
-            RawWaker::new(p, &VTABLE)
+        struct Noop;
+        impl std::task::Wake for Noop {
+            fn wake(self: std::sync::Arc<Self>) {}
         }
-        static VTABLE: RawWakerVTable = RawWakerVTable::new(clone, noop, noop, noop);
-        // SAFETY: the vtable functions ignore the data pointer entirely.
-        unsafe { Waker::from_raw(RawWaker::new(std::ptr::null(), &VTABLE)) }
+        std::sync::Arc::new(Noop).into()
     }
 
     #[test]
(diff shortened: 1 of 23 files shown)


crates/sigstore-crypto/src/hash.rs
```diff
@@ -24,6 +24,35 @@ pub fn sha512(data: &[u8]) -> Vec<u8> {
...
+        }
+    }
+
+    pub fn update(&mut self, data: &[u8]) {
+        self.context.update(data);
+    }
+
+    pub fn finalize(self) -> ArtifactDigest {
+        ArtifactDigest::new(self.algorithm, self.context.finish().as_ref())
+            .expect("cryptographic backend returned the algorithm's documented digest length")
```
jku: nit: error message should say what went wrong, not the opposite

jku (dismissed):
Some of this code I am not really able to review... but it seems correct to me and the concept certainly makes sense.

Having so many versions of verify/sign is a bit annoying but they all seem reasonable and serve a use case that's not covered by the other variants.

crates/sigstore-verify/examples/verify_bundle.rs
```diff
@@ -223,18 +224,17 @@ async fn main() {
...
+        let artifact = match fs::File::open(artifact_or_digest) {
+            Ok(file) => file,
             Err(e) => {
-                eprintln!("Error reading artifact '{}': {}", artifact_or_digest, e);
+                eprintln!("Error opening artifact '{}': {}", artifact_or_digest, e);
                 process::exit(1);
             }
         };
-        verify(&artifact_bytes, &bundle, &policy, &trusted_root)
+        verifier.verify_reader(artifact, &bundle, &policy)
```
baszalmstra: The example now unconditionally streams files. The reader API intentionally rejects Ed25519 message
signatures, whereas the previous byte API could verify them. This is code-traced, not reproduced with
an Ed25519 certificate fixture.

Could we preserve a byte-input path for schemes that cannot be prehashed? Switching every file to
`verify_reader` drops Ed25519 message-signature support. Those bundles could previously verify here,
and supplying a digest is not an alternative.

crates/sigstore-crypto/src/hash.rs
```diff
@@ -143,4 +296,156 @@ mod tests {
...
+        // itself, so a spin loop with a no-op waker is a sufficient executor.
+        let waker = futures_task_noop_waker();
+        let mut cx = TaskContext::from_waker(&waker);
+        let mut future = Box::pin(future);
+        loop {
+            if let Poll::Ready(output) = future.as_mut().poll(&mut cx) {
+                return output;
+            }
+        }
+    }
```
baszalmstra: Why not use `futures::executor::block_on`?

crates/sigstore-crypto/src/hash.rs
```diff
@@ -143,4 +296,156 @@ mod tests {
...
+
+    fn patterned(size: usize) -> Vec<u8> {
+        (0..size).map(|i| (i % 251) as u8).collect()
+    }
+
+    fn block_on<F: Future>(future: F) -> F::Output {
+        // Every future here only ever yields once per chunk and re-wakes
+        // itself, so a spin loop with a no-op waker is a sufficient executor.
+        let waker = futures_task_noop_waker();
```
baszalmstra: https://doc.rust-lang.org/stable/std/task/struct.Waker.html#method.noop?

crates/sigstore-sign/src/sign.rs
```diff
@@ -93,24 +70,34 @@ fn dsse_pae_hasher(payload_type: &str, payload_len: usize) -> Sha256Hasher {
...
+            self.hasher.update(data);
+        }
+    }
+    let mut prepared = CopyAndHash {
+        payload: Vec::with_capacity(data.len()),
+        hasher: dsse_pae_hasher(payload_type, data.len()),
+    };
+    hash_reader_yielding(data, std::slice::from_mut(&mut prepared))
+        .await
+        .expect("reading from an in-memory slice cannot fail");
```
baszalmstra: Could we share the yielding update logic without routing in-memory slices through a reader buffer?
This now copies each chunk into the 64 KiB buffer before copying it into `payload`. The previous chunk
loop hashed and copied directly from the source slice.

crates/sigstore-crypto/src/hash.rs
```diff
@@ -79,17 +136,113 @@ impl Default for Sha256Hasher {
...
+    loop {
+        match reader.read(buf) {
+            Err(e) if e.kind() == io::ErrorKind::Interrupted => continue,
+            result => return result,
+        }
+    }
+}
+
+/// Adapter presenting a blocking [`Read`] as an always-ready [`AsyncRead`].
+struct BlockingReader<R>(R);
```
baszalmstra, line 227: Maybe we can use https://docs.rs/futures/0.3.32/futures/io/struct.AllowStdIo.html ?

crates/sigstore-crypto/src/hash.rs
```diff
@@ -79,17 +136,113 @@ impl Default for Sha256Hasher {
...
-    let mut buf = [0u8; 8192];
+    hash_reader(reader, std::slice::from_mut(&mut hasher))?;
+    Ok(hasher.finalize())
+}
+
+/// Feed `reader` to EOF into every hasher in constant memory.
+///
+/// Reads block the calling thread. Inside an async task use
+/// [`hash_reader_yielding`] or [`hash_async_reader`] instead.
+pub fn hash_reader<H: HashUpdate>(mut reader: impl Read, hashers: &mut [H]) -> io::Result<()> {
```
baszalmstra, line 161: Could we expose reader hashing through the existing hasher types rather than adding more free
functions? A shared trait could provide `update_reader`, `update_reader_yielding`, and `update_async_reader`, with a hasher collection supporting the same contract for single-pass
multi-digest hashing. That would also remove the SHA-256 reader wrappers in `sign.rs`, while keeping
the one-shot functions as conveniences.

baszalmstra (commented):
Took a quick look. Overall looks good to me.

crates/sigstore-crypto/src/hash.rs
wolfv: I removed the SHA-256-specific reader wrappers from `sign.rs`, but kept traversal as free functions. These operations apply to a collection of hashers and include async I/O policy; putting them on the trait would require additional async-trait/future machinery under our Rust 1.70 MSRV. `HashUpdate` stays as the minimal shared contract for single-pass multi-digest hashing.

crates/sigstore-verify/examples/verify_bundle.rs
wolfv: Fixed in `fed1ee2`. Message-signature bundles now inspect the certificate’s signing scheme and retain the byte-input path when prehashing is unsupported, such as Ed25519. Other files, including DSSE artifacts, remain streamed. I added coverage for the selection logic.

crates/sigstore-sign/src/sign.rs
wolfv: Fixed in `fed1ee2`. In-memory data now uses `hash_yielding`, which hashes source slices directly while preserving the per-chunk yield. I added a test that verifies the chunks point into the original allocation.

crates/sigstore-crypto/src/hash.rs
wolfv: `Waker::noop` would be ideal, but it is newer than our Rust 1.70 MSRV. I replaced the custom implementation with `futures::task::noop_waker_ref` in the tests instead.

crates/sigstore-crypto/src/hash.rs
wolfv: I considered it, but kept the small adapter because `AllowStdIo` would require the full `futures` crate as a normal dependency; production code currently needs only `futures-io`. The adapter also preserves our explicit retry-on-`Interrupted` behavior.

crates/sigstore-crypto/src/hash.rs
wolfv: Good call—switched the tests to `futures::executor::block_on` and kept `futures` as a dev-dependency only.

baszalmstra (dismissed):

wolfv:
Rebased the stack root onto current `main`, resolved the verification-test conflict, and kept the new predicate-optional regression. CI is green; re-review requested because the branch update dismissed the previous approval.

baszalmstra (approved):

PR #193, opened 2026-08-19
refactor(types)!: deserialize semantic bundle values eagerly

wolfv:
Follow-up to #176. This moves validation to deserialization boundaries so downstream code works with semantic values rather than repeatedly parsing wire representations.

- deserialize bundle media types and supported Rekor kind/version pairs as enums
- parse checkpoints once while retaining their exact signed envelope bytes
- deserialize Rekor v1 proof hashes into fixed-size SHA-256 values (removing lossy fallback/filtering)
- represent in-toto SHA-256 and SHA-512 subject digests as fixed-size types
- reject malformed or unsupported values while parsing

The temporary base branch mirrors the exact #176 head so this PR shows only the follow-up diff. Once #176 merges, this PR should be retargeted to `main` and the temporary base deleted.

BREAKING CHANGE: bundle media types, Rekor kind/version pairs, checkpoints, proof hashes, and in-toto subject digests now use semantic types.

### DIFF
--- a/crates/sigstore-rekor/src/entry.rs
+++ b/crates/sigstore-rekor/src/entry.rs
@@ -89,11 +89,13 @@ pub struct RekorInclusionProof {
     /// Checkpoint (signed tree head)
     pub checkpoint: String,
     /// Hashes in the proof path (hex-encoded in V1 API)
-    pub hashes: Vec<String>,
+    #[serde(with = "hex_sha256_vec")]
+    pub hashes: Vec<Sha256Hash>,
     /// Log index
     pub log_index: u64,
     /// Root hash (hex-encoded in V1 API)
-    pub root_hash: String,
+    #[serde(with = "hex_sha256")]
+    pub root_hash: Sha256Hash,
     /// Tree size
     pub tree_size: u64,
 }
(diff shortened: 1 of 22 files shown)


crates/sigstore-verify/src/verify.rs
```diff
@@ -497,23 +499,27 @@ fn verify_dsse_artifact_binding(
     }
     let matches = match artifact {
         Artifact::Blob(bytes) => {
-            let sha256 = hex::encode(sigstore_crypto::sha256(bytes));
-            let sha512 = hex::encode(sigstore_crypto::sha512(bytes));
+            let sha256 = sigstore_crypto::sha256(bytes);
+            let sha512 = Sha512Hash::try_from_slice(&sigstore_crypto::sha512(bytes))
+                .expect("SHA-512 produces a 64-byte digest");
```
jku, line 504: I guess if we are changing things we could make sha512 return a Sha512Hash now that it exists -- this would match sha256...

crates/sigstore-rekor/src/entry.rs
jku, line 85: this looks at least partly misleading now

jku (approved):
This looks great

PR #199, opened 2026-09-05
fix(types)!: checked log indices, unambiguous digests and DSSE verifier binding

wolfv:
Stack: follows #198 (then #185).

- Make LogIndex construction fallible; propagate invalid untrusted Rekor indices instead of panicking.
- Require essential fields when constructing TlogEntryBuilder, removing its invalid default.
- Decode DigestBytes JSON as base64 only; provide an explicit from_hex constructor. Accept ProtoJSON base64 alphabets/padding for byte fields.
- Bind DSSE v0.0.1 logged verifiers to caller-supplied managed keys even when tlog inclusion verification is skipped, and check the payload hash algorithm.

Validation: workspace all-feature tests and strict all-target Clippy passed; regressions include a valid managed-key DSSE signature with correct, substituted and malformed logged verifiers.

Breaking: LogIndex::new returns Result; use TryFrom<u64> instead of From. TlogEntryBuilder::new requires entry fields. Hex digest JSON must be migrated to base64.

### DIFF
--- a/crates/sigstore-bundle/src/builder.rs
+++ b/crates/sigstore-bundle/src/builder.rs
@@ -133,7 +133,7 @@ impl BundleV03 {
 
 /// Helper to create a transparency log entry.
 pub struct TlogEntryBuilder {
-    log_index: u64,
+    log_index: LogIndex,
     log_id: String,
     kind_version: KindVersion,
     integrated_time: Option<jiff::Timestamp>,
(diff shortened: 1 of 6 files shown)


crates/sigstore-bundle/src/builder.rs
jku, line 238: for completeness this could take a LogIndex or `TryInto<LogIndex>` but that does not seem too important

crates/sigstore-types/src/encoding.rs
```diff
@@ -13,6 +13,27 @@ use crate::error::{Error, Result};
 use base64::Engine;
 use serde::{Deserialize, Serialize};
 
+// ProtoJSON bytes permit standard/URL-safe alphabets, with or without padding.
+fn decode_protojson_base64(value: &str) -> std::result::Result<Vec<u8>, base64::DecodeError> {
```
jku, line 17: this looks reasonable but it seems like we should be using it in several other places as well (but likely not quite every STANDARD.decode() location). Maybe file a followup issue?

jku (approved):
These changes seem a unrelated to each other, or maybe I'm missing the connection. I'd like to ask for more details on decode_protojson_base64() -- like why not use it in other places that do sigstore proto decoding -- but I don't want to prevent the other changes in this PR (and the rest of the stack) from merging...

Anyway, LGTM

crates/sigstore-types/src/encoding.rs
wolfv: Good catch — #223 updates the remaining ProtoJSON log-ID path and adds coverage for all four permitted base64 variants. I audited the other strict decoders; they handle non-ProtoJSON formats (signed checkpoints, JWTs, or Rekor’s extra signature layer), so I left those strict.

PR #200, opened 2026-09-05
fix(io): cancellable OAuth callbacks and atomic cache records

wolfv:
Stack: follows #199.

- Keep OAuth callback sockets async and bound the total request headers, so the existing deadline cannot be defeated by a partial request.
- Bound token exchange responses/deadlines and redact bearer tokens from Debug/error output.
- Check/remove expired memory entries under one lock.
- Replace filesystem expiration and data together using an existing workspace dependency (tempfile); never unlink a replacement during expiry reads. Reject overflowing TTLs.
- Default filesystem caches to the production namespace. Legacy two-file records are ignored.

Validation: all-feature tests and strict all-target Clippy for sigstore-cache and sigstore-oidc. Regressions cover partial/oversized callbacks, state mismatch, redaction, expiry and concurrent cache access.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2088,6 +2088,7 @@ dependencies = [
  "jiff",
  "serde",
  "serde_json",
+ "tempfile",
  "thiserror",
  "tokio",
 ]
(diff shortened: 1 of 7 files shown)


jku (approved):
This seem to be a collection of unrelated changes in two different crates but all seem good to me

wolfv:
Yeah I squashed the stacked PRs a bit to reduce review pressure :) Sorry about that

PR #201, opened 2026-09-05
fix(test): reject mutation-harness false positives and honor fixture trust

wolfv:
Stack: follows #200.

- Require successful unmodified verification before attempting mutations.
- Treat crashes, timeouts, usage/setup errors and unrecognized rejection diagnostics as harness errors, not passing mutations.
- Pass supplied trusted-root files to the Rust verifier.
- Add offline --trusted-root support to the verification example; custom TUF instances now require --tuf-root rather than silently trusting a writable cache.
- Add standard-library Python regression tests.

Validation: Python unittest checks, real offline baseline plus signature-bit mutation, and strict example Clippy. /usr/bin/false now exits the harness with status 2 rather than reporting every mutation passed. External verifier diagnostic changes intentionally require updating rejection markers.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,8 +151,13 @@ jobs:
       - name: Make binary executable
         run: chmod +x target/release/examples/verify_bundle
 
-      - name: Clone sigstore-conformance test data
-        run: git clone --depth 1 https://github.com/sigstore/sigstore-conformance.git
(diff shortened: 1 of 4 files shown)


jku (approved):
> Treat crashes, ..., as harness errors

I suppose this is still not ideal -- it's likely not a harness error -- but I agree this is a nice improvement as it's no longer categorized as "success"

PR #202, opened 2026-09-05
chore(release): tighten dependencies, MSRV and feature checks

wolfv:
Combines the dependency, MSRV and isolated-feature cleanup previously split across #202–204.

- Remove unused dependency declarations and narrow Tokio features.
- Repair TLS feature forwarding.
- Declare and test Rust 1.86 as the workspace MSRV.
- Check each crate with default, no-default and individual features.
- Make native-TLS dependency checks fail closed.

Validation: workspace tests, Rust 1.86 checks, strict Clippy, isolated crate feature checks and cargo-machete passed on the combined stack.

Stack: follows #201; #206 follows this PR.


jku (approved):
check-features is ~3 minute test... but probably no point in optimizing since we have several other 3 minute tests

LGTM

PR #206, opened 2026-09-05
docs(release): repair examples, license and trust-bootstrap guidance

wolfv:
Small follow-up to #202.

- Fix README snippets to use current APIs and explicit signer requirements.
- Align license/repository/version/MSRV documentation with Cargo and LICENSE; qualify FIPS claims.
- Require `--tuf-root` with custom instances in the signing example, matching the verification example.

Most changed files contain only the one-line license correction. Validation: examples, workspace tests, documentation build and Clippy passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,11 +31,13 @@ Add the crates you need to your `Cargo.toml`:
 ```toml
 [dependencies]
 # For verification only
-sigstore-verify = "0.1"
-sigstore-trust-root = "0.1"
+sigstore-verify = "0.11"
+sigstore-trust-root = "0.11"
+sigstore-types = "0.11"
 
 # For signing
-sigstore-sign = "0.1"
+sigstore-sign = "0.11"
+sigstore-oidc = "0.11"
 ```
 
 ## Usage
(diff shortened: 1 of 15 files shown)


jku:
this not targeting main currently

wolfv:
Now it does I think

README.md
```diff
@@ -53,25 +55,22 @@ let verifier = Verifier::new(&root);
...
-use sigstore_sign::{Signer, SigningConfig};
+use sigstore_sign::{SigningContext, SigningConfig};
 
-// Create a signer (will authenticate via OIDC)
-let config = SigningConfig::production();
-let signer = Signer::new(config).await?;
+let endpoints = sigstore_trust_root::SigningConfig::production().await?;
+let config = SigningConfig::from_tuf_config(&endpoints)?;
+let token = sigstore_oidc::get_identity_token(config.oidc_url.as_deref()).await?;
+let signer = SigningContext::with_config(config).signer(token);
```
jku: looking at this a couple of ideas (for future, not this PR:
* SigningConfig::production() really should default to being  async and backed by TUF
* we could have a  `SigninContext.authenticate()`: that could just use the oidc url from the config it has

then the whole oidc auth becomes:
```rust
let signer = SigningContext::production().await?.authenticate().await?;
let bundle = signer.sign(artifact_bytes).await?;
```
and with explicit token:
```rust
let signer = SigningContext::production().await?.signer(token);
let bundle = signer.sign(artifact_bytes).await?;
```

crates/sigstore-sign/examples/sign_blob.rs
```diff
@@ -304,7 +325,9 @@ fn print_usage(program: &str) {
     eprintln!("Options:");
     eprintln!("  -o, --output <FILE>  Output bundle path (default: <artifact>.sigstore.json)");
     eprintln!("  -t, --token <TOKEN>  OIDC identity token (skips interactive auth)");
-    eprintln!("      --instance <URL> Use a custom Sigstore instance");
+    eprintln!(
+        "      --instance <URL> --tuf-root <FILE>  Custom instance with trusted TUF bootstrap"
+    );
```
jku: I'm ok with this although I didn't do it this way on purpose originally: I believe this is kind of security theater -- we claim to be secure because the root file is now not cached **by us**... but reality is that user now stores the root file in ~/Downloads/ or somewhere just as secure.

jku (approved):
👍

PR #207, opened 2026-09-05
fix(crypto): validate and share algorithm resolution

wolfv:
Combines the crypto algorithm validation and resolver reuse previously split across #207–208.

- Validate SPKI algorithms, parameters, curves and BIT STRING encoding.
- Support checked P-384, RSA and ML-DSA autodetection alongside P-256 and Ed25519.
- Reuse the checked resolver for certificates, managed keys and TSA verification.
- Remove duplicate OID and curve parsing.

Validation: workspace all-feature tests and strict all-target Clippy passed, including P-384, RSA, ML-DSA and TSA fixtures.

Stack: follows #206; #209 follows this PR.

### DIFF
--- a/crates/sigstore-crypto/src/verification.rs
+++ b/crates/sigstore-crypto/src/verification.rs
@@ -1,20 +1,84 @@
 //! Signature verification using aws-lc-rs
 
 use crate::error::{Error, Result};
-use crate::signing::SigningScheme;
+use crate::signing::{KeyAlgorithm, SigningScheme};
 use aws_lc_rs::signature::{
     UnparsedPublicKey, ECDSA_P256_SHA256_ASN1, ECDSA_P256_SHA384_ASN1, ECDSA_P384_SHA256_ASN1,
     ECDSA_P384_SHA384_ASN1, ED25519, ML_DSA_44, ML_DSA_65, ML_DSA_87, RSA_PKCS1_2048_8192_SHA256,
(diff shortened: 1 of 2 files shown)


crates/sigstore-crypto/src/verification.rs
```diff
@@ -1,21 +1,85 @@
...
+impl KeyAlgorithm {
+    /// Resolve an SPKI algorithm with checked ASN.1 parameters.
+    pub fn from_spki(key: &DerPublicKey) -> Result<Self> {
+        let spki = SubjectPublicKeyInfoRef::try_from(key.as_bytes())
+            .map_err(|e| Error::InvalidKey(format!("Invalid SPKI: {e}")))?;
+        key_algorithm(&spki)
+    }
+}
+
+fn key_algorithm(spki: &SubjectPublicKeyInfoRef<'_>) -> Result<KeyAlgorithm> {
```
jku, line 29: the algorithm.oid matching in this method is now quite special  with multiple ifs and matches with other code in between them -- I'm assuming that some more structured approach is possible, maybe even a single `match algorithm.oid` would work here?

jku (dismissed):
Seems correct. left a comment on `key_algorithm()` but... looks like it's still correct

jku:
oh, this is also not against main branch

crates/sigstore-crypto/src/verification.rs
wolfv: Fixed this! :)

baszalmstra (approved):

wolfv:
Also filed an issue to use pre-defined OIDs once we can upgrade the crate (because `cms` is currently still holding it back ...).

PR #223, opened 2026-09-21
fix(types): accept ProtoJSON base64 log IDs

wolfv:
## Summary

- accept standard and URL-safe ProtoJSON base64 log IDs, with or without padding
- reuse the decoded log ID when constructing the SET payload

Follow-up to the review on #199. The remaining strict base64 decoders handle non-ProtoJSON formats (signed checkpoints, JWTs, and Rekor’s extra signature layer), so they are intentionally unchanged.

## Validation

- `cargo test -p sigstore-types -p sigstore-verify`


jku (approved):
Thanks!

